PR #14, opened 2025-12-26
Add XDG config file support for cargo install

ejc3:
## Summary

Makes rootfs config work for `cargo install` users by:
- Embedding config in binary via `include_str!()`
- Adding `fcvm setup --generate-config` to write config to XDG path
- Config lookup chain: `--config` → `~/.config/fcvm/` → `/etc/fcvm/` → next-to-binary

## User workflow

```bash
cargo install --git https://github.com/ejc3/fcvm
fcvm setup --generate-config   # Creates ~/.config/fcvm/rootfs-config.toml
fcvm setup                     # Downloads kernel, creates rootfs
```

## Test plan

- [x] Config generation works
- [x] Config lookup finds generated file
- [x] Helpful error when no config exists


ejc3:
Already merged to main

PR #32, opened 2025-12-27
Fix concurrent test failures with config file locking

ejc3:
## Summary
- Fix config file race condition that caused TOML parse errors when running tests concurrently
- Add disk space checks and cleanup targets to prevent disk-full failures

## Changes
1. **Config file locking** (`tests/common/mod.rs`):
   - Add file locking with `fs2::FileExt` to ensure only one process generates config
   - Other processes block on lock, then find config already valid (skip regeneration)

2. **Disk protection** (`Makefile`):
   - `check-disk`: Validates 10GB free on `/` and 15GB on `/mnt/fcvm-btrfs`
   - `clean-target`: Removes `target/` to free space on `/`
   - `clean-test-data`: Removes leftover VM disks, snapshots, state files
   - All test targets now depend on `check-disk` to fail fast

## Root Cause
The race condition occurred because:
- `std::sync::Once` is per-process, not cross-process
- cargo nextest runs each test in a separate process
- Multiple processes simultaneously writing config → partial reads → TOML parse errors

## Test Plan
- [x] Run `make test-root` with all tests passing concurrently
- [x] Verify `make check-disk` reports disk space correctly
- [x] Verify `make clean-test-data` cleans up leftover VM data


ejc3:
Superseded by #33 (Makefile changes only)

PR #34, opened 2025-12-27
Fix concurrent test config race with file locking

ejc3:
## Summary
Fix race condition when multiple test processes generate the config file simultaneously.

## Root Cause
- `std::sync::Once` (CONFIG_INIT) is per-process, not cross-process
- cargo nextest runs each test in a separate process
- Multiple processes write config simultaneously → partial reads → TOML parse errors

## Fix
- Add file locking with `fs2::FileExt` before config generation
- Processes block on lock until available
- Once locked, validate config before regenerating
- Lock released automatically on drop

## Test Results
222/227 tests pass (5 failures are unrelated: kvm, copy_file_range, fmt, clippy, exec_rootless)

The critical test: **No TOML parse errors** during concurrent test runs.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -1,6 +1,7 @@
 // Common test utilities for fcvm integration tests
 #![allow(dead_code)]
 
+use fs2::FileExt;
 use std::io::Write;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


ejc3:
Closing - superseded by later changes

PR #94, opened 2026-01-05
Improve setup flow: SUDO_USER config lookup and flexible btrfs paths

ejc3:
## Summary
- Add SUDO_USER config lookup so `sudo fcvm setup` reads config from the invoking user's home
- Change loopback path from hardcoded `/var/` to sibling of mount point
- Skip sudo if btrfs already exists at configured path
- Update README with generate-config step and manual btrfs instructions

## Changes
- `src/setup/rootfs.rs`: SUDO_USER lookup before $HOME in config search order
- `src/setup/storage.rs`: Loopback as sibling (e.g., `/mnt/fcvm-btrfs.img`)
- `src/commands/setup.rs`: Output says "sudo fcvm setup"
- `README.md`: Quick Start shows generate-config, alternative manual setup

## Test plan
- [x] Clean slate: generate-config → sudo setup → run container
- [x] Sanity tests pass (3/3)
- [x] Verified SUDO_USER lookup works

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -43,23 +43,43 @@ jobs:
             exit 1
           fi
 
+      - name: Check for non-member comments
+        id: check-comments
+        env:
+          GH_TOKEN: ${{ steps.claude-token.outputs.token }}
+        run: |
+          # Check if any non-member has commented (prompt injection risk)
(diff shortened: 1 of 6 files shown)


ejc3:
Replaced by #98 and upcoming security PR

PR #98, opened 2026-01-05
Improve setup flow: SUDO_USER config lookup and flexible btrfs paths

ejc3:
## Summary

- Look up config from invoking user's home when running with `sudo` (via SUDO_USER)
- Change loopback image path to be sibling of mount point (user-configurable)
- Skip root requirement if btrfs already exists at configured path

## Changes

- `src/setup/rootfs.rs`: Add SUDO_USER config lookup before $HOME
- `src/setup/storage.rs`: Loopback path follows mount point, early return if btrfs exists
- Fix `with_extension()` to append `.img` instead of replacing extension

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,14 @@ sudo iptables -P FORWARD ACCEPT
 # 1. Build
 cargo build --release --workspace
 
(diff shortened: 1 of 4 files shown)


ejc3:
dupe

PR #121, opened 2026-01-06
Improve Quick Start and fix rootless workflow

ejc3:
## Summary

- Improves README Quick Start with architecture diagram and complete setup instructions
- Makes rootless mode the default experience (no unnecessary sudo)
- Fixes Makefile to work on hosts where cargo is not in PATH
- Makes state files world-readable so `./fcvm ls` works without sudo

### README.md
- Add architecture diagram: `You → fcvm → Firecracker VM → Podman → Container`
- Add missing musl-tools and rustup target installation steps
- Use consistent `./fcvm` symlink throughout all examples
- Remove unnecessary sudo (only for setup, bridged, privileged)
- Reorganize: one-shot example first, then long-lived service
- Rename "Run a Container" to "More Options" to avoid duplication
- Add success indicators (`→`) showing expected output
- Make advanced demos use rootless mode for consistency

### Makefile
- Auto-detect Rust toolchain bin directory for hosts where cargo is not in PATH
- Prefer stable toolchain (has musl target) over nightly
- Use `$(CARGO)` variable throughout for consistency

### src/state/manager.rs
- Change state file permissions from 0600 to 0644 so non-root users can run `./fcvm ls`

## Test plan
- [ ] Verify Quick Start works on fresh host
- [ ] Verify `make build` works when cargo not in PATH
- [ ] Verify `./fcvm ls` works without sudo after state file permission change


ejc3:
@claude why did the build fail?

ejc3:
Fixed the security issue in commit 91ac3ea:

**Approach:** Instead of reverting to 0o600 (which would require sudo for `fcvm ls`), I removed the env var storage entirely.

**Why this is safe:**
- The `config.env` field was never read back from state
- Env vars are passed directly to MMDS at VM start time
- Clones use memory snapshots, not stored env vars
- The field was added in commit 444a467d as a general "store everything" pattern with no specific use case

This keeps the 0o644 permissions so non-root users can list VMs, while eliminating the secret exposure risk.

PR #137, opened 2026-01-09
Fix utimensat POSIX compliance and enable FUSE writeback cache

ejc3:
## Summary

Enables FUSE writeback cache for 9x write performance improvement, with a kernel patch to maintain full POSIX compliance.

## What Changed

**Sequence of changes in this branch:**

1. **Enable writeback cache** (`8c8f124`): 9x write performance, but breaks utimensat POSIX compliance
   - Disabled utimensat test as temporary workaround
   - Added `FCVM_NO_WRITEBACK_CACHE=1` env var for debugging

2. **Add kernel patch** (`003e635` → `90470ab`): Fixes utimensat with default_permissions
   - Patch fixes TWO issues in `fs/fuse/dir.c`:
     - `fuse_do_setattr()`: Add ATTR_FORCE when ATTR_TOUCH + write permission
     - `iattr_to_fattr()`: Send FATTR_MTIME_NOW when ATTR_TOUCH (even with writeback cache)
   - Re-enabled utimensat test

3. **Add tooling** (`881896b`): `scripts/kernel-patch.sh` to prevent hand-written patch format issues

## The Bug

With `default_permissions` + writeback cache, `utimensat(UTIME_NOW)` returns EPERM for non-owner users with write permission. POSIX says this should succeed.

**Root cause (two issues):**

1. `fuse_do_setattr()` only adds ATTR_FORCE when `!fc->default_permissions`. With default_permissions enabled, `setattr_prepare()` enforces owner-only, violating POSIX.

2. With writeback cache (`trust_local_cmtime=true`), `iattr_to_fattr()` sends FATTR_ATIME_NOW but NOT FATTR_MTIME_NOW. The FUSE server receives an explicit mtime timestamp, which requires ownership to set.

## Test Results

All 122 pjdfstest utimensat tests pass with the kernel patch applied.

```
$ make test-root FILTER=utimensat
✅ PJDFSTEST UTIMENSAT PASSED!
   All 122 utimensat tests passed with patched kernel.
```

## Test plan

- [x] Kernel builds with patch applied
- [x] pjdfstest utimensat category: 122/122 pass
- [x] `make lint` passes
- [x] `make test-root` passes


ejc3:
Closing - stale fix attempt, will fix directly in #133

PR #139, opened 2026-01-10
Fix O_WRONLY with writeback cache

ejc3:
## Summary

Fixes O_WRONLY files failing with EBADF when writeback cache is enabled.

**Stacked on:** fix-utimensat-kernel (PR #137)

## The Problem

`test_write_clears_suid` was failing with EBADF when opening files with O_WRONLY.
The previous "fix" was to disable the test with `#[ignore]` - this was wrong.

## Root Cause

fuse-backend-rs has `get_writeback_open_flags()` which promotes O_WRONLY → O_RDWR when writeback is enabled. But this only works when the internal `writeback` AtomicBool is set to true.

The AtomicBool is set in `init()` - and we weren't calling `init()`.

## The Fix

Call `inner.init(FsOptions::WRITEBACK_CACHE)` after creating PassthroughFs.

```rust
// Before: just import()
inner.import()?;

// After: init() with writeback capability
inner.init(FsOptions::WRITEBACK_CACHE)?;
```

## Changes

- `passthrough.rs`: Call init() with WRITEBACK_CACHE
- `test_permission_edge_cases.rs`: Remove `#[ignore]` - test now passes
- `CLAUDE.md`: Add "NEVER Assume" guideline

## Test Results

```
$ sudo cargo test --features privileged-tests \
    --test test_permission_edge_cases -- test_write_clears_suid
running 1 test
test test_write_clears_suid ... ok
```

## Test plan

- [x] test_write_clears_suid passes
- [x] No tests disabled/ignored


ejc3:
Changes cherry-picked into PR #133. This PR is now empty.

PR #192, opened 2026-01-26
fix: enable IP forwarding for podman container networking

ejc3:
## Problem

Container builds (like pjdfstest) hang indefinitely during `apt-get update` because the container can't reach external networks.

## Root Cause

When podman creates the `podman0` bridge interface, it inherits the default forwarding setting:
- `net.ipv4.conf.default.forwarding = 0` → `net.ipv4.conf.podman0.forwarding = 0`

Even though global `ip_forward=1`, the per-interface `forwarding=0` blocks packet routing through that interface.

**Packet flow that was broken:**
```
Container (10.88.x.x) → podman0 → [BLOCKED - forwarding=0] → enP1s33 → internet
```

## Fix

Enable forwarding for all interfaces BEFORE podman creates the bridge:
```bash
sudo sysctl -w net.ipv4.conf.all.forwarding=1
sudo sysctl -w net.ipv4.conf.default.forwarding=1
```

## Verification

On the broken runner:
```bash
# Before fix
$ ping -c 1 8.8.8.8  # from container
100% packet loss

$ sysctl net.ipv4.conf.podman0.forwarding
net.ipv4.conf.podman0.forwarding = 0

# After fix
$ sudo sysctl -w net.ipv4.conf.podman0.forwarding=1
$ ping -c 1 8.8.8.8  # from container
64 bytes from 8.8.8.8: icmp_seq=1 ttl=117 time=0.774 ms
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,6 +231,10 @@ jobs:
           sudo sysctl -w vm.unprivileged_userfaultfd=1
           # Disable AppArmor restriction on unprivileged user namespaces (needed for rootless networking on newer Ubuntu)
           sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0 2>/dev/null || true
+          # Enable IP forwarding for all interfaces (including future ones like podman0)
+          # This fixes podman container networking - without this, containers can't reach the internet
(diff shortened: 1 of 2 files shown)


ejc3:
Changes merged into PR #188 (diff-snapshot branch)

PR #197, opened 2026-01-31
fix: use podman HEALTHCHECK and restart health monitoring for clones

ejc3:
## Summary

Replace HTTP-based health checks with podman's built-in HEALTHCHECK mechanism and fix health monitoring for cloned VMs.

### Changes

**Health check refactor (commit 1):**
- Remove `health_check_url` field from all structs (VmConfig, NetworkConfig, SnapshotRunArgs)
- fc-agent monitors container health via `podman inspect --format '{{.State.Health.Status}}'`
- fc-agent sends "health:<status>" to host via vsock
- Host writes status to `container-health` file in VM runtime directory
- health.rs reads file instead of making HTTP requests
- Add HEALTHCHECK to Containerfile.nested for test image

**Clone health monitoring fix (commit 2):**
- fc-agent tokio tasks don't survive snapshot/restore cycle
- Add `restart_health_monitoring()` called from `handle_clone_restore()`
- Finds running container via `podman ps` and spawns new monitor task
- Uses existing restore-epoch detection mechanism

**Test fixes:**
- Increase ftrace test timeout to 120s for I/O contention scenarios
- Remove `--health-check` flag from all test invocations
- Remove obsolete `test_no_startup_snapshot_without_health_check_url` test

## Test plan

- [x] `make test-root FILTER=sanity` - all 3 tests pass (rootless, bridged, ftrace)
- [x] Fresh boot VMs become healthy via podman HEALTHCHECK
- [x] Clone VMs restart health monitoring and become healthy
- [x] Startup snapshots created after health check passes


ejc3:
Superseded by #200 which took a different approach (AND logic combining HTTP + podman HEALTHCHECK rather than replacing HTTP entirely).

PR #237, opened 2026-02-05
Enable clone_fd for true parallel FUSE request processing

ejc3:
## Summary
- Switch fuser dependency to `remap-file-range-on-clone-fd` branch (based on upstream PR #636)
- Add explicit `config.clone_fd = true` to opt-in to FUSE_DEV_IOC_CLONE for parallel requests

## Changes
- **fuser branch**: `remap-on-627` → `remap-file-range-on-clone-fd`
- **mount.rs**: Added `config.clone_fd = true` in both:
  - `mount_internal()` - Unix socket mounts (tests, local mounts)
  - `mount_vsock_with_options()` - vsock mounts (fc-agent in VMs)

## Why
The clone_fd feature uses FUSE_DEV_IOC_CLONE to create per-thread /dev/fuse file descriptors, enabling true parallel request processing instead of serializing through a single fd.

## Test plan
- [x] `cargo fmt --check` passes
- [x] `cargo clippy -p fuse-pipe --all-targets` passes
- [x] `cargo test -p fuse-pipe --lib` passes (42 tests)
- [x] Integration test passes (`cargo test -p fuse-pipe --test integration basic`)
- [ ] CI passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,9 +184,9 @@ checksum = "5dd9dc738b7a8311c7ade152424974d8115f2cdad61e8dab8dac9f2362298510"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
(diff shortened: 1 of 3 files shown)


ejc3:
Closing - clone_fd feature causes TTY hangs and test failures. Needs investigation before re-enabling.

PR #347, opened 2026-02-16
Add btrfs storage driver for rootless podman containers

ejc3:
## Summary

- When running localhost containers in rootless mode with `--user` (keep-id), podman's overlay driver hits `checkAndRecordIDMappedSupport()` which returns false for rootless, forcing `storage-chown-by-maps` (expensive chown-copy of every image layer). Btrfs handles user namespaces natively — no idmapped mounts, no chown fallback.
- fc-agent now auto-detects btrfs kernel support at runtime and configures podman to use a btrfs loopback filesystem instead of overlay
- New "btrfs" kernel profile provides a kernel with CONFIG_BTRFS_FS=y

### fc-agent (`fc-agent/src/agent.rs`, `fc-agent/src/container.rs`)
- Add `setup_btrfs_storage_if_available()` — detects `/sys/fs/btrfs`, creates 8G sparse loopback, mkfs.btrfs, mounts at `/var/lib/containers/storage`, writes storage.conf
- Add `setup_user_btrfs_storage()` — creates user-specific btrfs subdir with proper ownership for rootless podman
- Both functions called during agent startup, before any podman operations
- Fix: check `truncate` exit status using match pattern consistent with `mkfs.btrfs` and `mount`

### FirecrackerConfig (`src/firecracker/config.rs`)
- Add `user: Option<String>` field — ensures `--user` changes invalidate the snapshot cache key

### Rootfs & kernel config (`rootfs-config.toml`, `kernel/`)
- Add `btrfs-progs` to system packages
- Add `[kernel_profiles.btrfs.arm64]` and `[kernel_profiles.btrfs.amd64]` sections
- New kernel config fragments: `kernel/default.conf` (arm64), `kernel/default-x86.conf` (x86_64)

### CI (`.github/workflows/ci.yml`, `.github/workflows/kernels.yml`, `Makefile`)
- Build btrfs kernel during CI setup (both host and host-root jobs)
- Add `build-btrfs-kernel` job to kernels.yml for GitHub releases
- Add btrfs kernel build to Makefile `_setup-fcvm` target

### Tests (`tests/test_localhost_rootless.rs`)
- `test_localhost_rootless_btrfs_storage` — verifies btrfs storage driver is active
- `test_localhost_rootless_btrfs_keepid` — verifies rootless user namespace with btrfs works

PR #348, opened 2026-02-16
Derive MMDS plan from FirecrackerConfig, not CLI args

ejc3:
**Stacked on:** btrfs-local-container (PR #347)

## Summary

Make `FirecrackerConfig` the single source of truth for all user-input fields in the MMDS plan. Previously, `build_and_send_mmds()` read fields directly from CLI args, meaning a field could be in the MMDS plan without being in the snapshot cache key. `forward_localhost` was the concrete example: changing `--forward-localhost` reused the same stale snapshot.

### `src/firecracker/config.rs`
- Add `forward_localhost: Vec<u16>` to `FirecrackerConfig` (now part of snapshot cache key)
- Add `MmdsRuntime` struct for runtime-only MMDS values (proxies, timestamps, volume/disk assignments)
- Add `to_mmds_json(&self, runtime: MmdsRuntime)` method — builds MMDS JSON from config fields + runtime values
- Add `test_snapshot_key_changes_with_forward_localhost` test

### `src/commands/podman/vm_config.rs`
- Refactor `build_and_send_mmds()` to take `&FirecrackerConfig` instead of `&RunArgs`
- Remove `cmd_args` parameter (now read from config)
- Build `MmdsRuntime` from runtime locals (proxy resolution, volume ports, disk devices)
- Pass `forward_localhost` to `FirecrackerConfig::new()` call

### `src/commands/podman/snapshot.rs`
- Pass `forward_localhost` to `FirecrackerConfig::new()` call

### `.claude/skills/pr-workflow/SKILL.md`
- Add CI artifact download instructions for investigating test failures
- Add troubleshooting guide for CI not auto-starting (branch not based on main)

### DIFF
--- a/.claude/skills/pr-workflow/SKILL.md
+++ b/.claude/skills/pr-workflow/SKILL.md
@@ -109,6 +109,11 @@ Comments may contain:
 - **CI failure analysis** - Re-run if infra issue, fix if code issue
 - **Security concerns** - Must address before merge
 
+**Fix ALL severity levels.** Low-severity findings (naming inconsistencies, minor
+code quality issues, pattern mismatches) should still be fixed before merging.
+They're quick to address and prevent technical debt from accumulating. Only skip
+a finding if you have a concrete reason to disagree with it — not because it's low severity.
+
 **CRITICAL: Never Apply Skip Conditions!**
 
 When reviewing auto-fix PRs or suggested fixes:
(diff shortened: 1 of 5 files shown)


ejc3:

Stacked on #347. The top of the stack (#373) is green on all jobs.

PR #352, opened 2026-02-16
Fix rootless podman in VM: subuid passthrough, pipe deadlock, duplicate name detection

ejc3:
**Stacked on:** mmds-from-config (PR #348)

## Summary
- Pass host's `/etc/subuid` range through MMDS so VM user namespace matches the storage image UIDs built by rootless podman on the host
- Write `storage.conf` to `~/.config/containers/` for rootless mode (instead of `/etc/containers/`), chown `.config` so podman can create subdirs
- Fix pipe deadlock: `fuse_pipe=debug` flooded vsock with thousands of FUSE mux log lines/sec, filling the 64KB pipe buffer and blocking entrypoint stdout. Changed to `fuse_pipe=warn`.
- Detect duplicate VM names in `exec --name` and return error with PID list instead of silently picking one

## Tests added
- `test_parse_subid_file_finds_current_user`: verifies correct user lookup from temp subuid file
- `test_parse_subid_file_returns_none_for_missing_user`: graceful when user not in file
- `test_parse_subid_file_returns_none_for_missing_file`: graceful on nonexistent file
- `test_load_state_by_name_duplicate_detection`: errors with PID list for duplicate names, succeeds for unique names

## Test plan
- `make build` compiles cleanly
- VM boots with correct subuid range (`[fc-agent] subuid/subgid: 1879048192:65536`)
- Entrypoint no longer hangs from pipe deadlock
- `fcvm exec --name www` errors when multiple VMs share the name
- Unit tests pass: `cargo test parse_subid` and `cargo test state_manager`

### DIFF
--- a/fc-agent/src/agent.rs
+++ b/fc-agent/src/agent.rs
@@ -132,7 +132,12 @@ pub async fn run() -> Result<()> {
             .get("USER")
             .map(|s| s.as_str())
             .unwrap_or("fcvm-user");
-        let (username, _uid, runtime_dir) = container::create_vm_user(user_spec, desired_name);
+        let subuid_range = plan
+            .subuid_start
+            .zip(plan.subuid_count)
+            .or_else(|| plan.subuid_start.map(|s| (s, 65536)));
+        let (username, _uid, runtime_dir) =
+            container::create_vm_user(user_spec, desired_name, subuid_range);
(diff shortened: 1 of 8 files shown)


ejc3:

Stacked on #348. The top of the stack (#373) is green on all jobs.

PR #363, opened 2026-02-17
Native btrfs image build with correct UID ownership

ejc3:
**Stacked on:** vm-rootless-fixes (PR #352)

## Summary
- Three-mode image delivery: `--image-mode overlay|btrfs|archive` (auto-detected from kernel profile)
- Build btrfs storage images as the target user (uid 1000) via `runuser` so file ownership matches rootless podman expectations — no chown at boot
- Per-VM reflink copy of btrfs image (read-write, instant startup)
- Snapshot/restore support for btrfs mode

### Host side
- `ImageMode` enum replaces `--no-direct-image-mount`
- `build_btrfs_storage_image()` runs `podman load` as target user when `--user` specified
- Separate cache paths: `{digest}.btrfs.img` (root) vs `{digest}.btrfs-uid1000.img` (user)
- `mkfs.btrfs --rootdir --subvol` preserves btrfs subvolumes in image (requires btrfs-progs >= 6.12)
- Skip `runuser` when fcvm already runs as target uid (rootless without sudo)

### Guest side (fc-agent)
- Three-way dispatch on `image_mode` in agent.rs
- `mount_btrfs_for_user()`: mount at `~/.local/share/containers/storage` with correct ownership
- No runtime chown of btrfs content — files already correctly owned from host build
- `setup_btrfs_storage_if_available()` and `setup_user_btrfs_storage()` skipped for btrfs image mode

### Code quality (from auto-fix PR review)
- Deduplicate `User::from_name` calls in `storage_paths()`
- Merge `get_host_subuid_start/count` into single `get_host_subuid_range()` (was double-parsing `/etc/subuid`)
- Fix `test_parse_subid_file_finds_current_user` USER env var panic in CI
- cargo fmt + clippy fixes

## Test plan
- [x] 81 unit tests pass
- [x] 9/9 localhost E2E tests pass (run twice locally)
- [ ] CI green on all jobs

PR #369, opened 2026-02-17
Containerize btrfs image build for host compatibility

ejc3:
**Stacked on:** btrfs-native-image (PR #363)

## Summary
- Run `podman load --storage-driver btrfs` inside an Ubuntu Noble container instead of on the host, so it works on hosts without the btrfs podman driver (e.g. EL9)
- `build-btrfs-progs.sh` detects ARM64 vs x86-64 for the dynamic linker
- `cleanup_btrfs_tmp` uses recursive subvolume enumeration instead of depth-1 walk
- Restore `BTRFS_CACHE_VERSION` (v3) for cache invalidation
- Remove `.has-mount-program` marker from overlay storage images and rootfs
- `build_uid` uses numeric UID directly instead of `/etc/passwd` lookup
- Use whitelist cleanup (`clean_podman_state`) for btrfs storage, not blocklist
- Update README to reflect containerized build approach

## Test plan
- [ ] CI passes (lint, clippy, all test jobs)
- [ ] btrfs keepid permission issue needs separate investigation

### DIFF
--- a/scripts/build-btrfs-progs.sh
+++ b/scripts/build-btrfs-progs.sh
@@ -3,7 +3,9 @@
 #
 # btrfs-progs 6.12+ includes mkfs.btrfs --rootdir --subvol which creates
 # btrfs images from directories containing btrfs subvolumes — without root.
-# Ubuntu 24.04 ships btrfs-progs 6.6.3 which lacks this feature.
+#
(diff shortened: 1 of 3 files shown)


ejc3:

Stacked on #363. The top of the stack (#373) is green on all jobs.

PR #500, opened 2026-02-26
Remove custom pasta build, use distro package

ejc3:
## Summary
- Delete `scripts/build-passt.sh` and all references to building passt from source
- CI, AMI builder, Containerfile, and runner setup now use the distro-packaged `passt` from apt (already in the apt-get install lines)
- Update `scripts/preflight.sh` help message to suggest `apt-get install passt`

## Why
We don't use `--no-splice` or any custom pasta features — the distro package works fine. Building from source on every CI run adds unnecessary build time and complexity.

## Test plan
- [ ] CI passes (pasta is already in the apt-get install lines for all environments)
- [ ] Verify pasta binary is available in host and container CI jobs


ejc3:
Closing: distro passt (0.0~git20240220) is too old — can't open user namespaces inside containers (Permission denied). Custom build (2026_01_20.386b5f5) has ~2 years of fixes we depend on. The Debian source tarball approach on bake-snapshot-config solves the passt.top downtime without dropping to the broken distro version.

PR #514, opened 2026-02-27
Replace socat with built-in Rust TCP proxy

ejc3:
## Summary

Replace socat with a built-in Rust TCP proxy for routed networking mode. This eliminates the only external runtime dependency for data-plane proxying, making routed mode consistent with bridged (iptables) and rootless (pasta).

### Core change: `src/network/tcp_proxy.rs`

Namespace-aware TCP relay using `nix::sched::setns()` + `tokio::io::copy_bidirectional`:

- `connect_in_namespace(ns_path, addr)`: Enter network namespace, connect TCP, restore original namespace (guard pattern ensures restoration even on error)
- `bind_in_namespace(ns_path, addr)`: Same setns dance for bind+listen
- `start_port_forwards()`: Listen on host loopback IP, relay to guest inside namespace
- `start_proxy_relay()`: Listen inside namespace on gateway, relay to host proxy

### Changes

- **New**: `src/network/tcp_proxy.rs` — ~120 lines of namespace-aware TCP proxy
- **Modified**: `src/network/routed.rs` — Replace socat spawn/kill with tcp_proxy calls; remove socat crash detection (Rust bind fails immediately with proper error)
- **Modified**: `src/network/mod.rs` — Add `pub mod tcp_proxy`
- **Modified**: `tests/test_signal_cleanup.rs` — Remove socat PID assertions (proxy is in-process now)
- **Modified**: `tests/test_port_forward.rs`, `tests/test_snapshot_clone.rs` — Update comments
- **Modified**: `README.md`, `Containerfile`, `.github/workflows/ci.yml` — Remove socat from deps

### Additional fixes in this PR

- **`parse_proxy_addr`**: Support hostnames and trailing slashes (not just `ip:port`)
- **KVM dirty page tracking**: Detect ARM64 KVM failure to track dirty pages (returns all-ones bitmap), auto-retry as Full snapshot with clear error message
- **Partial port forward cleanup**: Abort previously-started proxy tasks if a later port forward bind fails
- **`ss -K` filter syntax**: Fix kill filter to use correct `sport = :port` syntax
- **fc-mock bridged tests**: Gate behind `privileged-tests` feature (require iptables/root)
- **Diff snapshot log file discovery**: Search for log files with flexible naming pattern
- **Remove startup snapshot fallback**: Don't silently fall back to fresh boot on snapshot failure
- **Snapshot type log**: Fix misleading "Diff" message after KVM diff-to-Full retry
- **Post-resume liveness check**: Fix stale comment about removed fallback behavior
- **Clone port forward verification**: Add loopback port forwarding check after clone restore
- **Forward-localhost timeout**: Increase nc timeout from 5s to 30s for CI stability
- **Clone port forward health check**: Add `--health-check http://localhost:80` to baseline VMs so health monitor validates nginx is serving before declaring healthy (fixes flaky clone port forward tests)
- **Per-VM snapshot lock**: Add flock-based exclusive lock in `create_snapshot_core()` to serialize concurrent snapshot operations on the same VM (fixes race between internal startup snapshot and external `fcvm snapshot create`)

## Test plan

- [x] `sudo make test-root FILTER=port_forward_routed STREAM=1` — TCP proxy port forwarding
- [x] `sudo make test-root FILTER=sanity_routed STREAM=1` — Basic routed VM lifecycle
- [x] `sudo make test-root FILTER=sigterm_cleanup_routed STREAM=1` — Cleanup without socat
- [x] `sudo make test-root FILTER=clone_port_forward_routed STREAM=1` — Clone port forwarding
- [x] `sudo make test-root FILTER=clone_port_forward_rootless STREAM=1` — Clone port forwarding (rootless)
- [x] `sudo make test-root FILTER=forward_localhost STREAM=1` — Forward localhost with 30s timeout
- [x] `cargo fmt --check && cargo clippy --all-targets -- -D warnings`


ejc3:
## CI Status

**12/13 checks passed.** Host-Root-x64-SnapshotEnabled passed (23m48s).

The single failure is **Host-Root-arm64-SnapshotEnabled** — this is a **pre-existing failure on main** (not caused by this PR):

| Test | Main (CI #22496153907) | This PR |
|------|----------------------|---------|
| `test_port_forward_routed` | FAIL (122s timeout) | FAIL (122s timeout) |
| `test_vsock_connect_lifecycle_stress_routed` | FAIL | FAIL |

Main's arm64-SnapshotEnabled job also fails with the exact same tests timing out at 122s. The x64 runner passes both tests cleanly on this PR, confirming the TCP proxy replacement works correctly.

PR #594, opened 2026-06-02
fix: post-merge green follow-ups (export warning, pasta netlink race)

ejc3:
Two fixes for the failures in the post-merge CI run on main (run 26827965718):

- The export digest re-verification added in the review follow-ups turned legitimate concurrent rebuilds of shared localhost tags into hard failures (test_localhost_archive_mode, test_localhost_default_mode, test_localhost_rootless_btrfs_keepid_ext4). It is now a warning naming both digests; exporting by an immutable reference remains the proper follow-up.
- The new pasta supervision attributed the recurring "Cannot find device pasta0" clone failures to pasta dying at startup with "netlink: Unexpected sequence number". Upstream fixed this after our pinned commit (d00255bd, netlink: use a regular request/response socket for the initial neighbour sync); that patch is now carried alongside the addr_seen patch, with the build directory keyed on the combined patch contents.

Verified: cargo fmt/clippy clean; both passt patches apply to the pinned tarball and the patched source builds; bash -n passes on the build script. The localhost-image and clone connection/stress suites in this PR's CI run are the end-to-end check.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -20,8 +20,8 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # Build passt from source for consistent version across environments
-COPY scripts/build-passt.sh scripts/passt-addr-seen.patch /tmp/
-RUN /tmp/build-passt.sh && rm -rf /tmp/passt-build* /tmp/build-passt.sh /tmp/passt-addr-seen.patch
+COPY scripts/build-passt.sh scripts/passt-*.patch /tmp/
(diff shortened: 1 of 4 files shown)


ejc3:
Agreed on the stale-digest cache entry: when the tag is rebuilt mid-export, the archive published under the old digest holds the newer content. That gap predates this PR (before #588 the export did the same thing silently; #588 turned it into a hard failure, this PR turns it into a warning so the race stays visible without failing legitimate parallel rebuilds). The real fix is exporting by the immutable reference captured at inspect time, tracked in https://github.com/ejc3/fcvm/issues/598.

PR #603, opened 2026-06-03
test: make e2e tests reliable under CI load

ejc3:
## Summary

Make the e2e suite reliable under CI load by fixing genuine test-reliability bugs (not by reducing parallelism or weakening assertions). Each commit fixes a distinct flake surfaced by CI:

1. **79c1fcd** — keep two e2e tests within their CI time and log budgets.
2. **1642e67** — fix Test 6's busybox `wget` egress check: drop `-q` (it suppressed the `-S` status headers) and `-6` (busybox wget doesn't support it); parse the `HTTP/...` status line, accepting 2xx/3xx (ecr returns 308).
3. **8611411** — `run_exec` now spawns with explicit `Stdio::piped()` + `kill_on_drop(true)` and a 120s host-side timeout, so a hung guest command can't consume the whole nextest budget or leave an orphan exec process.
4. **73a9767** — pace the non-blocking-output test slower (10ms→50ms) so best-effort delivery doesn't flake under CI load.
5. **992ecdc** — don't create a pre-start snapshot in the nested ficlone test: it's one-shot and never restores, and the snapshot pause/resume intermittently wedges vsock recovery on the NV2 nested path. Matches the sibling `test_copy_file_range_in_vm`.
6. **4d152f8** — same rationale for `test_exec_parallel_pipe_stress`: a one-shot exec-stress test that boots one VM, runs 100 execs, and kills it (never restores). In SnapshotEnabled CI the pre-start snapshot sits on its boot→healthy path; under peak concurrent disk contention the ~1GB memory dump took ~45s, blowing its `poll_health_by_pid(60)` budget before any exec ran (sole failure in run val2-x64se, 538/539). Add `--no-snapshot`; the snapshot path is covered elsewhere, and this also drops the VM's contribution to box-wide dump contention.

## Test
- `cargo build --release --tests` clean, `cargo clippy --release --all-targets` clean, `cargo fmt --check` clean.
- Validated together with the other SnapshotEnabled-stability fixes (#613, #615, #616) on a combined branch.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -510,6 +510,28 @@ pub async fn spawn_fcvm_with_log_path(
     args: &[&str],
     name: &str,
 ) -> anyhow::Result<(tokio::process::Child, u32, PathBuf)> {
+    spawn_fcvm_with_log_path_inner(args, name, true).await
+}
+
+/// Like `spawn_fcvm_with_log_path` but never echoes the child's output to the
+/// console — it only goes to the debug log file.
+///
+/// Use this for tests whose container deliberately emits a large volume of output
+/// (verified by polling the log file): echoing megabytes of container output into
+/// the CI job log can overwhelm the self-hosted runner's log uploader and drop the
+/// runner. The file still captures everything for assertions and debugging.
+pub async fn spawn_fcvm_with_log_path_quiet(
+    args: &[&str],
+    name: &str,
+) -> anyhow::Result<(tokio::process::Child, u32, PathBuf)> {
+    spawn_fcvm_with_log_path_inner(args, name, false).await
+}
+
+async fn spawn_fcvm_with_log_path_inner(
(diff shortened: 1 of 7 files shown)


ejc3:
Both P2 inline comments are already addressed in the current diff (they were filed against an earlier push):

- **Status-line parsing** (`tests/test_exec.rs` Test 6): the check no longer scans the whole header block. It filters to lines starting with `HTTP/` and parses the status token via `split_whitespace().nth(1)`, accepting only `2xx`/`3xx` (commit `1642e67`). A `Date:` header containing `" 2"` can't satisfy it.
- **kill-on-drop**: `run_exec` now spawns with explicit `Stdio::piped()` + `.kill_on_drop(true)` and a 120s `tokio::time::timeout`, so a hung `fcvm exec` is killed when the timeout fires rather than orphaned (commit `8611411`).

PR #605, opened 2026-06-03
build: align carried passt addr_seen patch with the upstream submission

ejc3:
Switches the carried scripts/passt-addr-seen.patch to the exact shape proposed upstream on passt-dev, so CI builds and exercises the code we actually submitted rather than a stricter local variant.

Changes vs the patch on main:
- Gate the addr_seen update on `no_dhcp`: the restriction to the configured guest address only applies when DHCP is disabled. fcvm always passes `-a`/`--no-dhcp`, so `c->no_dhcp` is set and behaviour is identical to before; with DHCP enabled (not fcvm) behaviour is unchanged, which is what makes it upstream-acceptable.
- Drop the IPv6 hunk: fcvm runs pasta with `--ipv4-only`, and the upstream patch is IPv4-only. Dead code for fcvm.

For fcvm this is a no-op behaviourally; it just keeps our build in lock-step with the upstream proposal.

## Test results
- Applies to the pinned 038c51e tarball with `patch -p1` and the patched tree builds (verified on a clean extract).
- build-passt.sh keys its build dir on the patch contents, so this forces a passt rebuild in CI.
- The clone port-forward stress and clone_http suites in this PR's CI run are the behavioural check: forwarding to a restored clone must still work without retries.

### DIFF
--- a/scripts/passt-addr-seen.patch
+++ b/scripts/passt-addr-seen.patch
@@ -1,47 +1,36 @@
-Restrict addr_seen updates to the configured guest address.
+Restrict addr_seen updates to the configured guest address (static deployments).
 
-pasta tracks the latest source address seen on the tap interface
-(addr_seen) and uses it as the target for inbound port forwarding. The
-update accepts any IPv4/IPv6 frame, so when the tap interface sits on a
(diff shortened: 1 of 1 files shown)


ejc3:
@codex review

PR #607, opened 2026-06-03
fcvm: log a shutdown-race exec stream close at debug, not ERROR

ejc3:
When an exec stream ends before an Exit message the command's outcome is unknown, so exec still exits non-zero. But this is the expected, benign terminal condition when an exec races VM/container shutdown — most visibly the health monitor's `podman inspect` healthcheck during teardown, which spawns a child `fcvm exec` that printed `ERROR fcvm: Error: exec connection closed before an exit status was received` on every clean shutdown. That ERROR is pure noise; the parent already records the failed inspect as a WARN.

Introduce a typed `ExecConnectionClosed` error and downcast it in `main`: still exit(1), but log at debug instead of ERROR. The error's Display is unchanged, so behaviour and the existing eof/truncation unit tests are intact.

## Test results
- cargo build/clippy clean
- cargo test commands::exec — 4 tests pass, including read_exec_responses_eof_without_exit_is_an_error and the truncated-line case (Display message preserved).

### DIFF
--- a/src/commands/exec.rs
+++ b/src/commands/exec.rs
@@ -299,15 +299,31 @@ pub async fn cmd_exec(args: ExecArgs) -> Result<()> {
 
     // Use binary framing for any mode needing TTY or stdin forwarding
     // JSON line mode only for plain non-interactive commands
-    if tty || interactive {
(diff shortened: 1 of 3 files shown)


ejc3:
@codex review

ejc3:
@codex review

PR #613, opened 2026-06-03
fix: don't re-read the snapshot-spanning status connection (5-min startup stall)

ejc3:
Fixes the intermittent SnapshotEnabled-only stall where a VM took ~5 minutes to report its container ready, blowing test budgets (e.g. test_ficlone, CI run 26862321867).

## Root cause (proven from the host code + the exact 300s boundary)
fc-agent sends each status message on a SEPARATE vsock connection (`send_status` opens a fresh one for `cache-ready`, then `ready`, then `exit`). The host status listener, after handling `cache-ready` on connection A and writing `cache-ack`, stayed in an inner loop doing a second `read_line` on A (300s timeout) waiting for `ready`. But `ready` arrives on a NEW connection B, which the single accept loop can't take while it's blocked reading A.

Connection A is held open across the pre-start snapshot's pause/resume. Normally A closes cleanly and the second read returns EOF at once, so the listener re-accepts and gets B fast (~43s healthy). Intermittently — worse on the NV2 nested kernel — the post-resume vsock fails to propagate A's close, so the second read blocks the full 300s. That is exactly the boundary observed: `cache-ack` at T, `ready` at T+300s.

## Fix
After sending `cache-ack`, flush and `break` to re-accept instead of re-reading A. fc-agent never sends more than one message per connection, so the listener should return to `accept()` for the next one. `ready` is then delivered immediately regardless of when A's EOF arrives. No timeout or parallelism changes.

## Test results
- cargo build --release and clippy clean.
- The SnapshotEnabled matrix jobs are the behavioural check (no more 300s status stalls); validating on this PR's CI run.

### DIFF
--- a/fc-agent/src/container.rs
+++ b/fc-agent/src/container.rs
@@ -869,44 +869,71 @@ pub fn notify_cache_ready_and_wait(
             Ok(_) => {}
         }
 
-        if let Some(revents) = poll_fds[0].revents() {
-            if revents.contains(PollFlags::POLLHUP) || revents.contains(PollFlags::POLLERR) {
-                // vsock reset from snapshot RESTORE. This happens on warm start:
-                // the host restored this VM from a cached pre-start snapshot, and
-                // VIRTIO_VSOCK_EVENT_TRANSPORT_RESET killed all connections.
-                eprintln!("[fc-agent] cache-ack connection reset (snapshot restore detected)");
-                return CacheResult::WarmStart;
-            }
-        }
+        // Drain readable data BEFORE acting on POLLHUP/POLLERR. Linux can report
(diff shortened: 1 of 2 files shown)


ejc3:
Fixed in 23b12f7. You're right: the host `break` closes the connection right after writing `cache-ack`, and Linux reports `POLLIN | POLLHUP` together, so the prior POLLHUP-before-read ordering would discard the buffered ack and misclassify a cold start as a warm restore.

`notify_cache_ready_and_wait` now drains readable data first — on a short read containing `cache-ack` it returns `ColdStart`; only an EOF / hangup with no buffered ack returns `WarmStart`. That's correct POSIX semantics (`POLLHUP` does not mean "no data to read") and makes the host's close-after-ack safe.

Issue #619, opened 2026-06-03
[tracking] fcvm e2e / SnapshotEnabled CI flakiness: landed fixes and next steps


ejc3:
Tracking issue for stabilizing the fcvm e2e suite, with focus on the SnapshotEnabled CI matrix (which runs the full ~539-test suite twice on the same self-hosted runner to exercise snapshot persistence). This documents what was root-caused and landed, what is still open, and the next steps.

## Landed on main (tip 2782667)

Each fix has a verified root cause, not a timing/parallelism workaround:

- **pasta netlink startup race** (#616): pasta's `nl_status()` dies with "Unexpected sequence number" when a route/neighbour notification (seq 0) races its setup request/response calls under heavy parallelism, so pasta exits before its TAP appears. `PastaNetwork::post_start` now retries start + wait-for-device 4x (kill dead pasta, 200ms backoff). Upstream `d00255bd` only fixed the neighbour-sync path; the addr_seen patch was also submitted upstream.
- **5-minute startup stall** (#613): the host status listener, after writing `cache-ack` on the cache-ready connection, did a second `read_line` on the same connection waiting for `ready` (which arrives on a new connection). The connection spans the pre-start snapshot pause/resume; intermittently its EOF was not propagated, so the read blocked the full 300s timeout. Fix: flush + `break` to re-accept.
- **cache-ack misclassification** (#613): `notify_cache_ready_and_wait` decided cold-vs-warm from poll flags before reading. Closing the connection right after `cache-ack` could make the guest see `POLLIN|POLLHUP` together (discarding a buffered ack) or, when guest write-probe bytes were left unread, a RST (`ECONNRESET` -> `Failed`, aborting container launch on every reset). Now data-driven: `ColdStart` iff `cache-ack` is read; any connection end without an ack (EOF, hangup, reset errno) -> `WarmStart`; only unexpected errors are fatal.
- **localhost image collisions** (#615): parallel localhost-image tests built the same tag and shared a content digest, so the digest-keyed cache conflated them. Each test now uses a unique image name AND unique CMD content.
[...]

PR #620, opened 2026-06-06
test: scope one-shot signal_cleanup tests out of snapshots (#618)

### LINKED ISSUE #618: One-shot tests intermittently time out under snapshot-create disk contention (SnapshotEnabled CI)
## Summary

In the SnapshotEnabled CI matrix, one-shot tests (boot a VM, do one action, kill it, never restore) intermittently time out waiting for the VM to become healthy. The VM's pre-start snapshot is created on the boot→healthy path, and under concurrent disk-write contention the ~1GB memory dump can take 45s or more, exceeding the test's health-wait budget. Whether a given one-shot test hits this depends on ordering (whether it is the first to create a snapshot for its key, vs restoring a snapshot another test already cached).

## Evidence

- `test_exec_parallel_pipe_stress` timed out at 60s (run val2-x64se): 45s gap between `pausing VM for snapshot` and `copying disk` while the VM was paused for the memory dump. Fixed by `--no-snapshot` (it never restores).
- `test_signal_cleanup::test_sigint_kills_firecracker_bridged` and `test_sigterm_cleanup_bridged` timed out at 122s on Host-Root-arm64-SnapshotEnabled (run 26875012361), while the same tests passed in 13s on Host-Root-x64-SnapshotEnabled (snapshot cache hit → fast restore). The arm64 instances were first-to-create the snapshot → slow create under contention → exceeded the 120s health wait.

The double-checked per-key snapshot lock is correct (only the first VM per key creates; the rest skip), so this is genuine disk-write contention across distinct keys plus same-key waiters blocking on the create, not a thundering-herd bug.

## Affected tests (currently disabled, see code)

- `test_signal_cleanup::test_sigint_kills_firecracker_bridged`
- `test_signal_cleanup::test_sigterm_kills_firecracker_bridged`
- `test_signal_cleanup::test_sigterm_cleanup_bridged`

Disabled with `#[ignore]` linking here so the SnapshotEnabled matrix is green; they test signal handling, not snapshots.

## Fix options

1. **Scope one-shot tests out of snapshots** (`--no-snapshot`): the established pattern for one-shot tests that never restore (`test_copy_file_range_in_vm`, `test_ficlone_cp_reflink_in_vm`, `test_exec_parallel_pipe_stress`). These tests do not assert snapshot behavior, so the pre-start snapshot is incidental overhead. This both fixes the timeout and reduces total dump contention. Preferred for the signal_cleanup tests.
2. **Bound concurrent snapshot memory-dumps**: a system-wide limit on how many VMs write a memory snapshot at once, so the disk is not saturated. Targets the contended resource directly without reducing test parallelism, but is a cross-process mechanism.

Re-enable the tests once one of the above lands (option 1 is a small change per test).

ejc3:
Closes #618.

## Problem
In the SnapshotEnabled CI matrix every `podman run` creates a pre-start snapshot on the boot→healthy path. The `signal_cleanup` tests boot a VM, send a signal, and assert cleanup — they **never restore** a snapshot, so that pre-start snapshot is pure overhead. Under concurrent disk-write contention the ~1GB memory dump can take 45s+, exceeding the test's health-wait budget when the VM is first-to-create a snapshot for its key. That timed out `test_sigint/sigterm_kills_firecracker_bridged` and `test_sigterm_cleanup_bridged` at 122s on arm64 SnapshotEnabled (they passed in 13s on a cache hit), so they were disabled with `#[ignore]` pending this fix.

## Fix
Add `--no-snapshot` to all six one-shot spawns in `test_signal_cleanup.rs` — the established pattern for boot-act-die tests (`test_proxy`, `test_exec`, `test_fuse_copy_file_range_vm`, …). These tests do not assert snapshot behavior, so skipping the pre-start snapshot both fixes the timeout and reduces total dump contention. This is issue #618's preferred option 1 (root-cause fix), not a timing workaround.

Removed the three `#[ignore = "...#618..."]` markers, re-enabling:
- `test_sigint_kills_firecracker_bridged`
- `test_sigterm_kills_firecracker_bridged`
- `test_sigterm_cleanup_bridged`

Applied `--no-snapshot` to the rootless and routed one-shot variants too (they also gain nothing from the pre-start snapshot). This proactively covers the #619 watch-item where `test_sigkill_kills_firecracker_rootless` flaked once under load — same one-shot-under-contention class.

## Test results
```
$ cargo fmt -p fcvm --check        # clean
$ cargo clippy --all-targets -- -D warnings   # clean
```
Behavioral validation is the SnapshotEnabled CI matrix on this PR — the exact environment where the timeouts reproduced.

### DIFF
--- a/tests/test_signal_cleanup.rs
+++ b/tests/test_signal_cleanup.rs
@@ -51,7 +51,6 @@ fn send_signal(pid: u32, signal: &str) -> Result<()> {
 /// correctly when running in parallel with other tests.
 #[cfg(feature = "privileged-tests")]
 #[test]
-#[ignore = "disabled pending #618: one-shot snapshot-create disk contention can exceed the health-wait budget in SnapshotEnabled CI"]
 fn test_sigint_kills_firecracker_bridged() -> Result<()> {
     println!("\ntest_sigint_kills_firecracker_bridged");
 
@@ -66,6 +65,11 @@ fn test_sigint_kills_firecracker_bridged() -> Result<()> {
             &vm_name,
             "--network",
             "bridged",
+            // One-shot test: boot, send a signal, verify cleanup. It never restores
+            // a snapshot, so skip the pre-start snapshot create (#618) — under
+            // SnapshotEnabled CI its ~1GB memory dump can exceed the health-wait
+            // budget when this VM is first-to-create under disk contention.
+            "--no-snapshot",
             common::TEST_IMAGE,
         ])
         .spawn()
@@ -189,7 +193,6 @@ fn test_sigint_kills_firecracker_bridged() -> Result<()> {
 /// correctly when running in parallel with other tests.
 #[cfg(feature = "privileged-tests")]
 #[test]
-#[ignore = "disabled pending #618: one-shot snapshot-create disk contention can exceed the health-wait budget in SnapshotEnabled CI"]
 fn test_sigterm_kills_firecracker_bridged() -> Result<()> {
     println!("\ntest_sigterm_kills_firecracker_bridged");
 
@@ -204,6 +207,11 @@ fn test_sigterm_kills_firecracker_bridged() -> Result<()> {
             &vm_name,
             "--network",
             "bridged",
+            // One-shot test: boot, send a signal, verify cleanup. It never restores
+            // a snapshot, so skip the pre-start snapshot create (#618) — under
+            // SnapshotEnabled CI its ~1GB memory dump can exceed the health-wait
+            // budget when this VM is first-to-create under disk contention.
+            "--no-snapshot",
             common::TEST_IMAGE,
         ])
         .spawn()
@@ -331,6 +339,9 @@ fn test_sigterm_cleanup_rootless() -> Result<()> {
             &vm_name,
             "--network",
             "rootless",
+            // One-shot test (boot, signal, verify cleanup); never restores a
+            // snapshot, so skip the pre-start snapshot create (#618).
+            "--no-snapshot",
             common::TEST_IMAGE,
         ])
         .spawn()
@@ -472,6 +483,9 @@ fn test_sigkill_kills_firecracker_rootless() -> Result<()> {
             &vm_name,
             "--network",
             "rootless",
+            // One-shot test (boot, signal, verify cleanup); never restores a
+            // snapshot, so skip the pre-start snapshot create (#618).
+            "--no-snapshot",
             common::TEST_IMAGE,
         ])
         .spawn()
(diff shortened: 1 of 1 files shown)


ejc3:
@codex review

ejc3:
@codex review

ejc3:
@codex the only failing job on this PR is `Host-Root-arm64-SnapshotEnabled`, where exactly one test failed: `test_snapshot_clone::test_snapshot_clone_stress_100_bridged` (FAIL at 153s, 537/538 passed). 

This PR only modifies `tests/test_signal_cleanup.rs` (adds `--no-snapshot` to the one-shot signal tests for #618, and a zombie-aware `process_running()` helper used only by `test_sigkill_kills_firecracker_rootless`). It does not touch `test_snapshot_clone.rs`, the snapshot/clone code paths, or any shared production code.

Please give an independent opinion: is this failure caused by this PR's changes, or is it a pre-existing load-dependent flake in the 100-clone bridged stress test on the contended arm64 SnapshotEnabled runner (the same disk-contention class tracked in #618/#619)? If you think it could be related to the PR, explain the mechanism.

ejc3:
Update: I reverted the zombie-aware `process_running()` commit from this PR. It made `test_sigkill_kills_firecracker_rootless` fail on **Host-x64** at the **holder** assertion ("namespace holder still running after fcvm SIGKILL"). This PR is now back to the pure #618 change (`--no-snapshot` on one-shot signal tests). The sigkill-cleanup robustness will be a separate, carefully-investigated PR.

@codex I'd like your independent opinion on the root cause of that holder failure, to make sure the follow-up fix is correct. Context:
- On SIGKILL of fcvm, its children should die via `PR_SET_PDEATHSIG` (set to SIGKILL). The namespace holder is `unshare --user --net -- sleep infinity`.
- The test tracks raw PIDs and checks `/proc/<pid>` presence. My reverted change treated a zombie (`/proc/<pid>/stat` state `Z`) as dead.
- The failure: the holder PID was reported "still running" at the final assertion, but the test completed in ~9s (the 15s poll did not run to completion), implying the poll loop saw it gone and broke early, then the assertion saw a process at that PID again.

Two hypotheses — which do you find more likely, and how would you fix it robustly?
1. **PID reuse**: the holder is reaped, its PID is recycled by another parallel test process between the poll-break and the assertion, so the bare PID check sees an unrelated process. Fix: verify identity (e.g. `/proc/<pid>/stat` start-time, or `/proc/<pid>/comm`/cmdline) before declaring "still running".
2. **Holder survives SIGKILL**: `unshare --user` creates a new user namespace, and a credential change zeroes `task->pdeath_signal` — so if fcvm sets the holder's pdeathsig *before* the user-ns is created, it's cleared and the holder (`sleep infinity`) is orphaned, not killed. Fix: ensure the holder's pdeathsig is (re)established after the user-ns, or have fcvm track+kill the holder another way on SIGKILL.

Please assess both and recommend the correct fix.

Issue #617, opened 2026-06-03
Exec into a restored VM (2nd SnapshotEnabled pass) hangs to the 600s ceiling


ejc3:
## Summary

`test_readme_examples::test_custom_command_bridged` hangs and TIMEOUTs at 600s, but only on the **second** SnapshotEnabled pass (the snapshot-persistence pass that restores the snapshots created by the first pass). The first (cold) pass passes. The test execs into the container after the VM is healthy; that exec is what hangs.

## What the log shows (run 26875012361, Host-Root-x64-SnapshotEnabled, VM `cmd-test-base-3340047`)

```
09:22:28  Pre-start snapshot hit! Restoring from cached snapshot   (2nd pass restore)
09:22:28  cache-ack connection reset on read (warm start); detected restore-epoch; handling restore
09:22:28  exec server: vsock transport reset -> re-registered -> exec re-registered after restore
09:22:28  restore complete; nginx start worker processes; GET / HTTP/1.1 200; VM healthy
09:22:28  signaling startup snapshot trigger; Creating startup snapshot (diff)
09:22:28.984  pausing VM for snapshot
09:22:29.870  copying disk (VM paused)
09:22:29.954  VM resumed, processing snapshot
09:22:30.462  Startup snapshot created successfully
   ... no further VM output ...
09:32:24  TIMEOUT [600.004s] test_custom_command_bridged
```

The test body: boot with `--cmd` (writes a marker then runs nginx) and `--health-check`, wait healthy, then `exec_in_container(["cat", "/tmp/marker.txt"])`, then `exec_in_container(["wget", ...])`. The first exec is where it hangs.

## What this rules out

The startup snapshot is **not** the cause and does **not** break the vsock. It completed in ~1.5s, the VM resumed, and nginx kept serving (HTTP 200). A snapshot *create* (pause → dump → resume) leaves the running VM and its vsock intact; only a *restore* resets the transport, and that restore's exec-server re-register is logged as successful here.

(The earlier "snapshot-create pause/resume disrupts vsock, only restore triggers rebind" framing was wrong and has been removed.)

## What is still unknown

Why the host-side `fcvm exec` (cat /tmp/marker.txt) never completes on the restore path, even though the restore re-registered the exec listener and the VM/container are healthy. Not yet determined whether the host `exec` connect hangs or the guest exec server's `accept()` never fires for this connection. The exec helper has no host-side timeout, so a hang shows up only as the 600s test ceiling.

## Hypothesis (unverified)

[...]

PR #622, opened 2026-06-06
test+agent: host-side exec timeout and guest accept trace for #617

ejc3:
Addresses #617 (step 1 of its plan). Does **not** close it.

## Background
`test_readme_examples::test_custom_command_bridged` hangs to the 600s nextest ceiling on the **second** SnapshotEnabled pass (the restore pass). Tracing the code: the host `fcvm exec` CONNECT handshake succeeds at Firecracker's vsock proxy (`OK <port>`), then the host blocks reading command output under a **3600s** internal read timeout while the guest exec server never services the connection. The test helper has no timeout, so the hang surfaces only as the 600s ceiling — which hides the cause.

The guest re-registers its vsock listener on restore (`exec_rebind_needed`, set only by the MMDS restore-epoch watcher → `VsockListener::re_register()`). Whether `accept()` then fires for a connection arriving *after* re-register (and after the startup snapshot that immediately follows the restore) is exactly what's unknown. The issue's plan: **make the failure legible first, then capture the guest `accept()` trace** to root-cause the real bug.

## This PR (no behavior change to production paths)
- **`tests/common/mod.rs`** — `exec_in_vm` / `exec_in_container` now run `fcvm exec` under a 120s host-side timeout (`kill_on_drop`), failing fast with a `#617`-referencing message instead of hanging to the ceiling (also satisfies #619 step 6: helper timeouts everywhere). Both delegate to a shared `exec_with_timeout`.
- **`fc-agent/src/exec.rs`** — log `accepted connection on vsock port N` when `accept()` fires. With the existing `re-registered` log, the serial console will now show whether `accept()` ever fires for a post-restore exec, distinguishing "listener never delivers readiness" from "hang downstream in `handle_connection`".

## Not done here (deliberately)
`test_custom_command_bridged` stays `#[ignore]`'d. Re-enabling a still-hanging test would only re-red the SnapshotEnabled matrix. It re-enables once the captured trace confirms the actual vsock fix.

## Test results
```
cargo fmt -p fcvm -p fc-agent --check    # clean
cargo clippy --all-targets --all-features -- -D warnings   # clean
```

cc @codex review — please sanity-check the timeout/kill_on_drop helper and the guest log placement.

### DIFF
--- a/fc-agent/src/exec.rs
+++ b/fc-agent/src/exec.rs
@@ -75,6 +75,15 @@ pub async fn run_server(
             result = listener.accept() => {
                 match result {
                     Ok(client_fd) => {
+                        // Diagnostic for #617: confirms accept() actually fired after a
+                        // restore. If a restored-VM exec hangs and this line is absent
+                        // from the serial log while "re-registered" is present, the
+                        // re-registered listener is not delivering readiness for new
+                        // connections (vs. the hang being downstream in handle_connection).
+                        eprintln!(
+                            "[fc-agent] exec server: accepted connection on vsock port {}",
+                            vsock::EXEC_PORT
+                        );
                         tokio::spawn(handle_connection(client_fd));
                     }
                     Err(e) => {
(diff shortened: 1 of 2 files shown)


ejc3:
@codex review

PR #623, opened 2026-06-07
fix: export localhost images by immutable image ID (#598)

### LINKED ISSUE #598: Export localhost images by an immutable reference
Follow-up from #588 and #594. fcvm resolves a localhost/ tag at inspect time and later runs podman save on the tag. If the tag is rebuilt in between (parallel test builds do this), the exported archive can hold newer content than the digest used as the cache key, so the cache entry for the old digest no longer matches its key. #594 downgraded the mismatch from a hard failure to a warning so the race is visible without failing legitimate runs; the real fix is to export by the immutable reference captured at inspect time (image ID or digest) and keep the archive loadable by name in the guest (an ID-only archive has no RepoTags). Needs a regression test that rebuilds the tag during export. Raised by Codex review on #594.

ejc3:
Closes #598.

## Problem
fcvm resolves a `localhost/` tag's digest at inspect time (`get_image_identifier`, the snapshot/image cache key) and runs the export later. The export used `podman save <tag>` — the **mutable tag** — so a parallel build that repointed the tag in between (parallel tests routinely rebuild shared localhost images) made the archive hold newer content than the digest its cache entry is keyed by. #594 downgraded that to a non-fatal warning; this is the real fix.

## Fix
Capture the **immutable image ID** alongside the digest at inspect time, and export with:
```
skopeo copy containers-storage:<image_id> docker-archive:<tmp>:<repo_tag>
```
- The image ID names exact content → the archive always matches the digest cache key, even if the tag is rebuilt mid-run.
- skopeo's `docker-archive` destination writes the original repo tag into the archive's `RepoTags`, so the guest still `podman load`s and runs it by name (no guest change needed).
- **Non-destructive**: unlike `podman tag <id> <tag>`, it never mutates the caller's live image store, so it can't clobber a concurrent rebuild.

The post-export digest re-check (the #594 warning) is removed: content is now pinned to the cache-key image, so the mismatch it warned about can no longer occur.

### Changes
- `src/commands/podman/image.rs`: `get_localhost_image_id()`; `export_image_archive()` (skopeo copy by id + manifest.json validation), re-exported from `podman`.
- `src/commands/podman/mod.rs`: capture the image id at inspect time, export via the new function.

## Test results (run locally, pass)
```
✅ export pinned v1 content across tag rebuild (#598)
PASS [0.552s] fcvm::test_localhost_image test_export_pins_immutable_content_across_tag_rebuild
```
Builds v1, captures its id, rebuilds the tag to v2, exports by the v1 id, and asserts the archive holds v1's `Cmd` (not v2) and preserves the repo tag.

cc @codex review — please scrutinize the skopeo `containers-storage:`→`docker-archive:` export, the image-id capture timing vs. the cache key, and the removal of the post-export digest re-check.

### DIFF
--- a/src/commands/podman/image.rs
+++ b/src/commands/podman/image.rs
@@ -106,10 +106,17 @@ pub(super) async fn get_image_cache_ref(image: &str) -> Result<ImageCacheRef> {
         )
     })?;
 
-    let cache_key = digest.trim().trim_start_matches("sha256:").to_string();
     let image_id = id.trim().to_string();
     if image_id.is_empty() {
         bail!("podman returned an empty image id for '{}'", image);
+    }
+
+    // A locally-built image can report an empty manifest digest. Fall back to the
+    // (always-present, immutable) image id so distinct images never collide on an empty
+    // cache key — which would conflate their cached archives/snapshots.
+    let mut cache_key = digest.trim().trim_start_matches("sha256:").to_string();
+    if cache_key.is_empty() {
+        cache_key = image_id.trim_start_matches("sha256:").to_string();
     }
 
     Ok(ImageCacheRef {
--- a/src/commands/podman/image.rs
+++ b/src/commands/podman/image.rs
@@ -78,6 +78,79 @@ pub(super) async fn get_image_identifier(image: &str) -> Result<String> {
     }
 }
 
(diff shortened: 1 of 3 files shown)


ejc3:
@codex review

ejc3:
@codex review

PR #631, opened 2026-06-07
diag: log uncovered vmstate-embedded rootfs dir at restore (#608)

ejc3:
Addresses #608 with evidence + a safe diagnostic (supersedes the closed #621).

## Empirical finding (refutes the original hypothesis)
#608 guessed, from a code read, that a startup/diff snapshot records the base's vm_id, dropping the embedded disk vm_id from `baseline_dirs`. I reproduced the multi-level chain locally and inspected raw `vmstate.bin`:
- Fresh VM `vm-af77…` → snapshot embeds `…/vm-disks/vm-af77…/disks/rootfs.raw` (own id).
- Clone `vm-7f3d…` (cache-restored from it, `patch_drive`) → snapshot embeds **`…/vm-disks/vm-7f3d…/disks/rootfs.raw`** — the clone's **own patched path**, not the base's.

So `CreateSnapshot` serializes the **patched (creator)** drive path ⇒ the embedded rootfs id is always the creating VM's id ⇒ `build_snapshot_config` (always sets `original_vsock_vm_id`) + restore (`snapshot_vm_id = Some(vm_id)`) **always** put it in `baseline_dirs`. The metadata always covers it; the divergence #621 fixed doesn't occur (which is why #621's test passed without the scan — Codex was right). #621 closed.

## This PR (log-only, no behavior change)
At restore, after `baseline_dirs` is built, scan `vmstate.bin` for the embedded `…/disks/rootfs.raw` path(s) and:
- `debug!` when the dir is covered by `baseline_dirs` (the normal/empirical case),
- `warn!` (with the embedded id + `baseline_dirs`) if it is **not** covered.

The real ~0.7% `LoadSnapshot … No such file` failure was real but its cause is undetermined and CI captured no evidence. This turns the next occurrence into self-diagnosing evidence instead of a guess. `baseline_dirs` and `set_mount_redirects` are unchanged. The scanner is anchored to `rootfs.raw` so an external `--disk` under a vm-disks dir isn't matched.

## Tests
- `test_rootfs_disk_vm_ids_in_vmstate` (run locally, pass): dedup, external-disk ignored, missing-file.
- fmt + clippy clean; `/code-review`: no blocking findings.

cc @codex review

### DIFF
--- a/src/commands/common.rs
+++ b/src/commands/common.rs
@@ -712,6 +712,56 @@ pub struct RestoreParams<'a> {
     pub track_dirty_pages: bool,
 }
 
+/// Diagnostic helper (#608): the `vm-disks/<id>` directory ids whose **rootfs** path is
+/// embedded in a Firecracker `vmstate.bin`.
+///
+/// Firecracker serializes each drive's `path_on_host` as a plain UTF-8 string, so we can
+/// read which rootfs path `LoadSnapshot` will open. Anchored to the `…/disks/rootfs.raw`
+/// suffix so an external `--disk` that merely lives under a vm-disks dir is not matched.
+/// Best-effort: an unreadable file or no matches yields an empty vec. Used log-only at
+/// restore to detect (and make diagnosable) any case where the embedded rootfs dir is not
+/// covered by `baseline_dirs` — empirically that never happens (CreateSnapshot serializes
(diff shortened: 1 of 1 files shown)


ejc3:
@codex review

Issue #632, opened 2026-06-07
[epic] Hypervisor-agnostic abstraction: support Firecracker AND Cloud Hypervisor (the two microVMs) behind a pluggable trait


ejc3:
## Goal
Refactor fcvm so the VMM is **pluggable**: the same fcvm features work on **Firecracker** (today) and **Cloud Hypervisor** (new), behind one clean `Hypervisor` trait. Backends declare **capabilities**; the orchestration layer degrades gracefully where a backend lacks a feature. No behavior change for the Firecracker path.

Why Cloud Hypervisor: longer-running workloads, UEFI/ACPI boot of stock images, VFIO/GPU passthrough, CPU/mem hotplug, device flexibility — while staying a Rust, KVM, security-focused microVM. fcvm should not be locked to one VMM.

---

## 1. Current Firecracker coupling (internal map)
~6,000 of ~28,740 LOC (~21%) touch Firecracker, concentrated in 4 areas:

- **`src/firecracker/` (4 files, ~1,440 LOC)** — the natural trait boundary:
  - `api.rs` `FirecrackerClient`: Unix-socket HTTP wrapper. Endpoints: `PUT /boot-source`, `/machine-config` (vcpu, mem, `track_dirty_pages`, `huge_pages`), `/drives/{id}` (+ `PATCH` retarget), `/network-interfaces/{id}`, `/mmds` + `/mmds/config`, `/vsock`, `/snapshot/create` (Full|Diff), `/snapshot/load` (`mem_backend: File|Uffd`), `/actions` (InstanceStart), `PATCH /vm` (Paused|Resumed).
  - `config.rs` `FirecrackerConfig`: VM config + `snapshot_key()` (cache key, includes `firecracker_bin`) + `to_mmds_json()` (the guest boot plan).
  - `vm.rs` `VmManager`: process spawn (api-sock, pre_exec user/mount/net namespaces, **bind-mount redirects**, `PR_SET_PDEATHSIG`), socket-readiness poll, teardown.
- **Snapshot/restore (`src/commands/common.rs`, ~350 LOC)** — `create_snapshot_core` (Full/Diff + dirty-page-loss detection/retry), `restore_from_snapshot` (LoadSnapshot→patch_drive ordering), `merge_diff_snapshot` (SEEK_DATA/HOLE).
[...]

PR #633, opened 2026-06-07
docs: hypervisor-agnostic abstraction design (Firecracker + Cloud Hypervisor)

ejc3:
RFC design doc for the pluggable-VMM epic (#632), refocused on **the two microVMs (Firecracker + Cloud Hypervisor)** — QEMU dropped as out of scope (it was the outlier: no host vsock UDS proxy, QMP vs REST, larger attack surface, microvm x86-only, ~3× boot).

Backed by deep research (internal coupling map + cited FC/CH/QEMU capability mappings in #632). Summarizes: the proven FC↔CH mapping, the `Hypervisor` trait + `Capabilities` + `GuestChannel` design, the four capability gates, the **CoW (not UFFD) basis for clone memory efficiency** (corrected), the MMDS→vsock boot-plan, why we don't adopt Kata's trait, risks/angles, and a P0–P3 plan.

@codex review — please do a **deep architectural review of the plan itself**, not just doc nits. Specifically challenge:
1. Is the `Hypervisor` trait the right seam/granularity, or will snapshot/restore + UFFD leak through it? Is `configure/start/snapshot/restore/guest_channel` sufficient and minimal?
2. The **capability-gate** approach vs. alternatives — does graceful degradation (e.g., CH clones without dense memory-share) create surprising UX/correctness traps?
3. The claim that **cross-clone memory sharing is CoW (MAP_PRIVATE file-backed), not UFFD-exclusive**, and that CH can replicate it via `--memory file=`/`shared=off` — is that sound, and is the restore-path caveat correctly identified?
4. **MMDS→vsock boot-plan** as the portable metadata path — any pitfall (ordering, restore re-delivery, security of the plan over vsock)?
5. The **single `GuestChannel` (UdsConnect)** assumption for FC+CH — any divergence we're missing (CID conventions, restore vsock-reset)?
6. The **snapshot format/version cache-key** gating — sufficient to prevent cross-backend/version restore corruption?
7. Phasing/sequencing risks (P0 trait extraction; P0.5 fc-agent boot-plan); anything that should move earlier/later.

Goal of this PR: get Codex's independent critique of the design before any implementation lands.

### DIFF
--- a/docs/hypervisor-abstraction.md
+++ b/docs/hypervisor-abstraction.md
@@ -1,0 +1,122 @@
+# Design: hypervisor-agnostic abstraction (Firecracker + Cloud Hypervisor)
+
+Status: **proposal / RFC** (epic: #632). Canonical design doc; #632 holds the full proven
+capability mapping with source citations.
+
+**Scope: the two microVMs — Firecracker and Cloud Hypervisor.** QEMU is explicitly out of
+scope (it was the outlier on every hard axis: no host vsock UDS proxy, QMP vs REST, a much
+larger attack surface, `microvm` x86-only, ~3× boot / ~2× RSS). Restricting to the two lean
+Rust/KVM microVMs makes the abstraction small and clean.
+
+## Goal
+Make fcvm's VMM **pluggable** behind one `Hypervisor` trait so the same features work on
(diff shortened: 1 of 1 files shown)


ejc3:
@codex review

Issue #625, opened 2026-06-07
Disk-only clone: cold-boot fresh VMs from a captured disk (free hugepage/memory choice)


ejc3:
## Original request (verbatim)
> "i want functionality that can clone a vm but not its memory state. this would let me do something like get a vm booted and then we clone the disk and then can start fresh copies from that cloned disk state. think about what the interface with that would be and come back with a plan"
>
> "i want the ability to start that fresh clone with huge tb's even if the original one wasnt started that way"
>
> (huge tb's = hugepages / huge-TLB memory)

And as success criteria the requester added:
> "make sure you run /code-review and codex adversarial review on the new feature as part of success"
> "need to also make sure documentation is coherent with new feature too"

## What this is (and how it differs from existing snapshot/clone)
fcvm's current `snapshot create`/`serve`/`run` is a **memory + disk** clone: it captures `vmstate.bin` + `memory.bin` + a disk reflink, and `snapshot run` *resumes* the VM mid-execution via a UFFD memory server (clones share physical pages). Optimized for instant warm clones.

This feature is a **disk-only** clone: capture ONLY the disk at a point in time, then **cold-boot** brand-new VMs from it. Each fresh VM boots its kernel from scratch, fc-agent re-launches the container, and there is **no shared memory, no UFFD, no vmstate** — fully independent copies that start from the prepared filesystem state (installed packages, cached image layers, files written before capture).

**Why hugepages "just works" on the clone (the second ask):** because a disk-only clone **cold-boots with freshly allocated memory** (no memory image to restore), the memory backing is decoupled from the source VM. The clone can request `--hugepages` (and any `--memory`/`--cpu`) freely. The existing constraint ("hugepages require UFFD; backing must match the snapshot", `src/commands/snapshot.rs:1001/1016`) applies only to *memory restore* and does not apply here.

## Agreed interface (requester chose: extend the snapshot commands)
```
fcvm snapshot create --pid <vm_pid> --tag base --disk-only          # capture disk only
fcvm snapshot run --tag base --name c1 --hugepages --memory 8192 --cpu 4   # cold boot
fcvm snapshot run --tag base --name c2                              # another fresh copy
```
- `--disk-only` snapshot stores only `disk.raw` (+ `:ro` extra disks) + `config.json` with `kind=DiskOnly`; **no** `memory.bin`/`vmstate.bin`; **no** `serve` step.
[...]

PR #650, opened 2026-06-09
docs: harden disk-only-clone RFC — clean abstraction + drop capture pause (#625)

ejc3:
Hardens the disk-only-clone design doc (docs/disk-only-clone.html) across multiple adversarial review rounds (local codex gpt-5.5 xhigh + /code-review fan-out), each finding verified against current `main` before encoding. Design/RFC only — no implementation (that lands via the P1→P4 stack).

## Highlights
- **Clean abstraction**: one capture-side policy (QuiescePolicy) + three boot-side Preserve-vs-Fresh policies over categories of captured guest state — StoragePolicy, ContainerStatePolicy, **IdentityPolicy** (now first-class) — distilled to *"preserve the work, regenerate the identity."* Plus a completeness argument: three categories suffice because a disk-only clone is a cold boot from a crash-consistent disk, so the policies are exactly the deltas a normal reboot can't fix.
- **Capture correctness**: drop the vCPU pause entirely (fsfreeze gives consistency; disk-only skips Firecracker create_snapshot), removing the resume-failure wedge and making the guest auto-thaw watchdog actually work.
- Data-loss/correctness fixes: tight runtime-lock cleanup (never db.sql/overlay-*), StartCaptured recovery (podman container cleanup → start --attach, TTY+non-TTY), identity regen in initrd before switch_root (cmdline-gated, incl. dbus machine-id), reject non-root --user + external --disk, no container_cmd persistence (secret), false-healthy gate, fail-closed on pre-feature state.
- ~50 file:line citations corrected to the actual command/construct.


ejc3:
Superseded by #651 (merged to main 2026-06-09). This branch hardened the disk-only-clone RFC (policy-based abstraction) BEFORE implementation. #651 then implemented a simpler design (guest self-detects via a /var/lib/fcvm/provisioned marker — no StoragePolicy/ContainerStatePolicy, no MMDS plumbing) and merged the AS-BUILT docs for README.md, DESIGN.md, and docs/disk-only-clone.html, including an implementation-outcome callout that records why the RFC's policy enums were dropped. Merging this now would conflict with and regress those docs back to the pre-implementation policy design. The RFC reasoning is preserved in the design doc's history (PR #648) and referenced by the callout on main.

PR #699, opened 2026-08-05
Add a 6.16.1 A/B kernel pair for mountinfo propagation cost

ejc3:
Reading /proc/<pid>/mountinfo in a slave mount namespace calls
get_dominating_id() for every slave line, and that scans the whole peer
ring of the mount's master. With a large mount table the scans dominate:
on the host this came from, one 985-line listing walks 258,589 ring
steps, and a single 508-member ring accounts for 99.8 percent of them.

These two profiles differ only by a patch, so the pair can be booted and
A/B'd:

  mountinfo-base  6.16.1, unpatched control
  mountinfo-fix   6.16.1 plus kernel/patches-mountinfo/

The patch memoises the first-level ring scan across one listing, keyed on
the master's peer group id. The key has to be the group id and not the
master pointer: turning a shared tree into slaves gives every slave its
own master, so the same listing presents 1,859 distinct master pointers
naming only 462 rings. Replaying the measured sequence, an 8-entry cache
leaves 2,049 of the 258,589 steps and 16 entries only reach 2,025.

Both profiles pin the same kernel_version and kernel_config so the patch
is the only difference between them.

## Summary by CodeRabbit

* **New Features**
  * Added kernel support for PCI, virtio-MMIO, FUSE, nested virtualization, networking, compressed SquashFS, and NFS.
  * Added mountinfo performance profiles for comparing baseline and optimized kernel behavior.
  * Improved mountinfo performance by caching repeated peer-group lookups.

* **Testing**
  * Added dedicated control and optimized profiles to support mountinfo performance A/B testing.


ejc3:
Thanks, two of these were real and would have made the comparison meaningless. Fixed in cbf6608.

**Folded — neither profile would have built 6.16.1.** `KernelProfile::is_custom()` requires both `kernel_version` and `kernel_repo` (`src/setup/rootfs.rs:209`), so with only the version set `inherits_kernel()` was true and both profiles fell back to the default kernel. Both now declare `kernel_repo = "ejc3/fcvm"`, matching `nested` and `btrfs`.

**Folded — the pair differed by more than the patch under test.** `patches_dir = None` is not "no patches": `src/setup/kernel.rs:637` maps it to `kernel/patches`, so the control would have taken the two FUSE patches while the fix profile took only the mountinfo one. `kernel/patches-mountinfo` now holds symlinks to every `kernel/patches` entry beside the patch under test, and the control names `kernel/patches` explicitly. Symlinks rather than copies so the baseline cannot drift between the two.

**Declined — the duplicate table declarations are not there.** `mountinfo-base` appears once at line 328 and `mountinfo-fix` once at 337, and the file parses:

```
$ python3 -c "import tomllib;d=tomllib.load(open('rootfs-config.toml','rb'));print(list(d['kernel_profiles']))"
['nested', 'btrfs', 'mountinfo-base', 'mountinfo-fix']
```

**Declined for this PR, but the observation is right.** `scripts/kernel-patch.sh` `get_patches_dir()` does ignore the profile's `patches_dir` and only returns `kernel/patches-arm64` or `kernel/patches`. That script is the helper for creating, editing and validating patches, not the build path, and the build reads `patches_dir` in `src/setup/kernel.rs`. So this PR's A/B is unaffected, but the helper would mis-resolve if pointed at this profile. Worth its own change.

**Not yet done:** the two kernels have not been built and booted, so there is no timing here. What the patch rests on so far is a model of work removed: replaying the measured sequence of master peer-group ids from a real 985-line listing, weighted by ring size, gives 258,589 ring steps uncached and 2,049 with an 8-entry cache.

ejc3:
Ran the A/B these profiles were built for. Both kernels come from the same 6.16.1 source and the same config, differing only by `kernel/patches-mountinfo/0003-fs-cache-dominating-id-per-peer-group.patch`.

First, proof the pair differs as intended and only there:

```
$ strings -a vmlinux-mountinfo-fix-6.16.1-x86_64-013b84c05623.bin  | grep -c get_dominating_id_cached
1
$ strings -a vmlinux-mountinfo-base-6.16.1-x86_64-bb750bd6f7f2.bin | grep -c get_dominating_id_cached
0
```

Both contain `get_dominating_id`; only the variant has the cached entry point.

Each kernel boots in a microVM, makes N bind mounts shared, then times reads of `/proc/self/mountinfo` inside a slave namespace. Medians of seven reads:

| namespace | unpatched | patched | faster by |
|---|---|---|---|
| 2,012 lines, 2,001-member peer group | 44.3 ms | 20.2 ms | **2.19x** |
| 312 lines, 301-member peer group | 4.86 ms | 4.06 ms | 1.20x |

At the larger size the two sets do not overlap: slowest patched read 21.6 ms, fastest unpatched 43.2 ms. The smaller size is the control on the mechanism rather than a second data point, since a cost of lines times peers should shrink with the mount table, and it does.

Two earlier attempts produced no valid measurement and are worth recording, because both looked like clean runs. The first reported zero slave mounts, so there was no propagation walk at all and both kernels read in 5-6 ms, which reads as "the patch does nothing". The second timed the wrong namespace: plain `unshare -m` defaults to private, so `mount --make-rslave /` yields no masters, while the `--propagation slave` namespace that did have 512 slaves was never timed. The guest script now aborts if fewer than 100 slaves are present, so a null setup fails loudly instead of returning a number.

The third failure was a quoting one rather than a mount one: an inline `bash -c` payload had its nested `$(...)` mangled somewhere between the host shell and the guest, so a `M=$(grep -c master: ...)` assignment read 0 while an identical substitution inside an `echo` read 312. The payload is now base64-encoded, which removes the question.

ejc3:
@coderabbitai review

PR #701, opened 2026-08-05
fix: honor --config and profile patches_dir when resolving kernels

ejc3:
Four lookups ignored what the caller configured, and each failed in the same confusing way: fcvm reported a thing as missing while pointing at a file or a binary that has it.

**`--config` now applies to every config load.** It previously reached only the call sites that threaded it explicitly, so anything going through `load_plan()` (kernel profile lookup among them) silently read the XDG user config. Passing `--config` at a file containing a profile still produced `kernel profile 'x' not found in config`. The path is recorded once in a `OnceLock` that `find_config_file()` consults, matching how `paths.rs` already handles data and assets directories, so all six `load_config(None)` call sites agree.

**The recording happens before the `--install-host-kernel` early return.** Placed after it, that path kept resolving profiles against the discovered config.

**A missing `--config` is an error, not a fallback.** The first version checked `if p.exists()` and fell through when it did not, reintroducing the silent-wrong-file behaviour. It now bails, like the explicit-argument branch above it. Supersedes #703.

**The rootfs setup VM uses the configured Firecracker.** That path called `Command::new("firecracker")`, a bare PATH lookup ignoring both `find_firecracker()` and `FCVM_FIRECRACKER_BIN`, while the same file's doc comment said otherwise. On a host with no system-wide Firecracker it failed with a bare ENOENT two lines after reporting its own Firecracker ready.

**`scripts/kernel-patch.sh` reads the profile's `patches_dir`.** `get_patches_dir()` only ever returned `kernel/patches-arm64` or `kernel/patches`, so for a profile declaring its own directory the helper created, edited and validated a different patch set than the build applies. Its matcher also required the table name to end at the profile, which no arch-suffixed table does, and `grep -A20` spans both arch blocks. Both it and `get_kernel_version()` now scope to the current arch's table, and an empty `patches_dir` means "apply none", as the Rust side already means by it.

**Testing.** `cargo test --release --lib`: 10 + 184 + 68 passed. Two new tests, both watched failing first:

- `setup_vm_firecracker_bin_honors_env_var` fails on the old bare-PATH resolver, naming itself and `firecracker not found in PATH`, and passes with the fix.
- `find_config_file_rejects_a_missing_explicit_path` covers the fallthrough.

Why no test caught the Firecracker one: the resolver is only reached while building a rootfs, and a cached rootfs skips it, so an end-to-end run proves nothing about it on a second invocation.

Found while building the mountinfo A/B kernels for #699.

## Summary by CodeRabbit

- **New Features**
  - Added support for explicitly selecting a configuration file during setup, with clear errors when the path is invalid.
  - Added support for selecting the Firecracker binary through an environment variable, with validation and automatic PATH fallback.
  - Setup now consistently uses the selected Firecracker executable.

- **Bug Fixes**
  - Architecture-specific kernel and patch settings are now read from the correct profile.
  - Empty patch directory settings can explicitly disable patch application.
  - Patch creation, editing, and completion now report when patches are disabled.
  - Configuration paths are honored across all setup flows.

### DIFF
--- a/scripts/kernel-patch.sh
+++ b/scripts/kernel-patch.sh
@@ -43,8 +43,12 @@ get_kernel_version() {
 
     # Extract kernel version for the profile
     # This is a simple grep - for complex configs, use a proper TOML parser
-    grep -A20 "\[kernel_profiles\.$profile\]" "$config_file" | \
-        grep -m1 'kernel_version' | \
+    # Same arch scoping as get_patches_dir: profile tables are arch-suffixed, and
+    # a bare -A20 spans both the .amd64 and .arm64 blocks.
+    local carch="amd64"
+    [[ "$(uname -m)" == "aarch64" ]] && carch="arm64"
+    sed -n "/^\\[kernel_profiles\\.$profile\\.$carch\\]/,/^\\[/p" "$config_file" | \
+        grep -m1 '^kernel_version' | \
         sed 's/.*=.*"\([^"]*\)".*/\1/' || \
         error "Could not find kernel_version for profile '$profile'"
 }
--- a/scripts/kernel-patch.sh
+++ b/scripts/kernel-patch.sh
@@ -220,6 +220,9 @@ cmd_create() {
     local patches_dir=$(get_patches_dir "$profile")
     local workdir="/tmp/kernel-patch-$$"
 
+    # An empty patches_dir means the profile deliberately applies no patches.
+    [[ -z "$patches_dir" ]] && error "Profile '$profile' has patches disabled (patches_dir is empty); cannot create a patch"
+
     info "Creating patch for kernel $version (profile: $profile)"
 
     # Setup kernel source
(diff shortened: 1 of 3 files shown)


ejc3:
Good catch, that was a real hole in the fix. The recording sat after the `--install-host-kernel` early return, so that path kept resolving profiles against the discovered config. Moved to the top of `cmd_setup`, ahead of every early return.

Verified both directions with a profile that exists only in the passed file:

```
$ fcvm setup --install-host-kernel --config /tmp/probe-config.toml --kernel-profile probeonly-xyz
Building and installing host kernel with profile 'probeonly-xyz'...

$ fcvm setup --install-host-kernel --kernel-profile probeonly-xyz
ERROR fcvm: Error: kernel profile 'probeonly-xyz' not found in config
```

The second is the control: without `--config` the same command still fails, so the check can go red.

ejc3:
Good catch, that was a real hole in the fix. The recording sat after the `--install-host-kernel` early return, so that path kept resolving profiles against the discovered config. Moved to the top of `cmd_setup`, ahead of every early return.

Verified both directions with a profile that exists only in the passed file:

```
$ fcvm setup --install-host-kernel --config /tmp/probe-config.toml --kernel-profile probeonly-xyz
Building and installing host kernel with profile 'probeonly-xyz'...

$ fcvm setup --install-host-kernel --kernel-profile probeonly-xyz
ERROR fcvm: Error: kernel profile 'probeonly-xyz' not found in config
```

The second is the control: without `--config` the same command still fails, so the check can go red.

Issue #708, opened 2026-08-06
Firecracker fork: serial device restore loses in-flight TX state — restored guests get permanently dead console


ejc3:
## Summary

A snapshot captured while the guest 8250 UART has TX bytes in flight produces a snapshot whose **every restore has permanently dead serial-console output**: the guest driver waits for a TX-complete interrupt that the re-created Firecracker serial device never delivers. fc-agent's own restore doc already notes "the UART IER register is stale" after restore (`fc-agent/src/restore.rs:31-39`) for the kernel printk path; this issue is the userspace-tty variant, and it is what actually broke CI.

## Evidence (CI run 31066211429 attempt 1, Host-Root-arm64-SnapshotEnabled)

- Creation race: host logged `Creating pre-start snapshot snapshot_key=c5d56efda67b` at 04:12:23.192684; the guest's `[fc-agent] sent cache-ready:… waiting for ack…` line surfaced on the host only at 04:12:24.773566 — 36µs before "VM resumed" — proving those bytes were captured undelivered in the memory image.
- Blast radius was deterministic, not flaky: **26/26** restores of that snapshot (and every UFFD clone chained from them) had zero serial output while their vsock channels reconnected fine; **199/199** restores of other snapshots had live serial. Perfect A/B at the same second: estress-bridged (other key) restored with serial; estress-rootless (poisoned key) restored without.
- Downstream, dead serial wedges fc-agent entirely: fd1/fd2 are dup2'd to /dev/console, the tty's 4096-byte buffer fills, and the next `eprintln!` blocks the exec-server accept loop → exec + health checks dead → `test_serial_console_after_restore`, `test_exec_rootless`, `test_exec_parallel_tty_stress` all failed both nextest tries and got written off as "transient infrastructure".

## fcvm-side mitigations (separate PR, in progress)

1. fc-agent drains the console (TIOCOUTQ → 0) before signaling cache-ready and stays quiet until resume, so the *pre-start* snapshot can never capture mid-TX state.
2. fc-agent console writes become non-blocking drop-on-full so a dead console degrades to lost log lines, never a wedged exec server.
3. Host logs a loud ERROR naming the snapshot when a restore's vsock reconnects but no serial line arrives.

## Why the fork fix is still needed

[...]

PR #718, opened 2026-08-07
Add deterministic VM lifecycle fuzzing: failpoints + three test tiers

ejc3:
Deterministic VM lifecycle fuzzing: a failpoint crate plus three test tiers that turn lifecycle races (snapshot pause / restore / Healthy publication colliding with in-flight exec, curl, and serial traffic) into replayable, individually named tests instead of probabilistic CI flakes.

This is the suite that found the exec-orphan, healthy-before-pause, and serial-wedge bug trio now fixed on main (#710, #711, #712); it lands after those fixes so every test passes deterministically.

## Why

This week's CI flakes were all one race class: a lifecycle transition (VM pause for snapshot, snapshot restore, first-Healthy persist) landing inside an in-flight guest/host interaction. The windows are microseconds wide, so stress testing hits them probabilistically and `retries=1` hid them inside green runs. This PR makes each window a first-class, widenable, sequenced test target.

## The failpoint crate (`failpoint/`)

- **Named hold points** compiled into fcvm and fc-agent. Unarmed (the only state production ever sees) a hit is one static lookup returning `None` — zero cost, no I/O.
- **Arming**: host via `FCVM_FAILPOINT` (read once at startup); guest via `FCVM_GUEST_FAILPOINT`, which fcvm forwards onto the kernel cmdline as `fcvm_failpoint=<spec>` (the `fuse_trace_rate` pattern) because fc-agent cannot read host env; fc-agent parses `/proc/cmdline` at boot.
- **Actions**: `<name>:sleep:<ms>` and `<name>:block_until_file:<path>` (host-only; the harness can't create files inside the guest, and a file-blocked guest would hold VM-global state). Both are hard-capped at 60s so a wedged harness can never wedge a VM. Guest specs are validated host-side (`validate_guest_spec`) before a VM boots; malformed specs panic loudly — a silently un-armed failpoint would make a determinism test vacuously pass.
- **Marker-line contract**: `FAILPOINT <name> reached action=...` before the hold, `FAILPOINT <name> released` after (plus a loud `RELEASED BY TIMEOUT`). Harnesses sequence on these exact boundaries instead of guessing timing — every Tier 1 test waits for `reached` before driving the other side of the race.
- **Hold points** (each names the race it makes testable): `snapshot.pre_pause`, `snapshot.post_save_pre_resume`, `restore.pre_resume`, `health.pre_persist_healthy` (host); `cache_ready.pre_send`, `exec.post_accept_pre_read`, `exec.post_ack_pre_go` (guest).

### Why guest failpoint specs feed the snapshot cache key

Guest failpoints ride the kernel cmdline, so the armed spec is baked into the guest's boot — and therefore into any snapshot taken of it. The runtime boot-args string is deliberately excluded from the snapshot cache key, so `FirecrackerConfig` gains a dedicated `guest_failpoint` field (populated from `FCVM_GUEST_FAILPOINT`, `skip_serializing_if` when `None`) that feeds the key. A fuzz VM with failpoints armed computes a different key than a normal run: fuzz VMs can never pollute normal snapshot caches, and a normal run can never restore a snapshot whose guest has failpoints armed. The existing golden-key test (`test_snapshot_key_golden`) still passes, proving keys for normal runs are byte-identical to before.

## Three tiers

**Tier 0 — deterministic protocol fuzz** (`fc-agent/src/exec_fuzz_tests.rs`, `src/commands/exec_fuzz_tests.rs`, fc-mock parity in `tests/test_fc_mock.rs`): scripted fake peers enumerate connection death at every ACK/GO handshake stage and byte offset, against the real agent handshake reader and the real client (`connect_and_start_exec` + `EpochGuardedReader` + `read_exec_responses`, exactly the `run_exec_in_vm` path). Includes the exactly-once property enumerated across all cut points: executions ≤ 1 always, == 1 whenever the client reports success. 21 tests, event-driven (server closes/shutdowns as cut events), no sleeps-as-sync.

**Tier 1 — lifecycle interleave matrix** (`tests/test_lifecycle_interleave.rs`, 5 tests): named interleavings with strict oracles, marker-sequenced:
- `exec_resend_across_agent_stall` — snapshot create held at `pre_pause` (block_until_file), exec parked in the agent's post-accept window, pause released into it: pre-ACK orphan must resend and the nonce must appear exactly once (double-execution tripwire).
- `go_floor_across_ack_stall` — 12s guest hold between ACK and GO (longer than the agent's 10s shared handshake deadline): only the recomputed 2s GO floor lets the buffered GO be consumed; one handshake attempt, no resend.
- `healthy_means_live_dataplane` — `snapshot.pre_pause:sleep:3000` inflates the startup-snapshot pause; at the FIRST observed Healthy a one-shot 2s curl must succeed (Healthy may never be visible while vCPUs are paused).
- `quiesce_survives_long_prepause` — guest hold before cache-ready + 2s host pre-pause hold, then a second (restored) run must have a live serial console.
- `no_leaks_after_interleaved_teardown` — SIGTERM into a held pause window; the pause must never complete (`released` marker absent) and process/state leak oracles must come back clean.

**Tier 2 — seeded chaos** (`tests/test_fuzz_chaos.rs`): schedule-reproducible op sequences (exec-nonce / tty-exec / port curl / state-read with Healthy⇒live chained curl / snapshot-create / console-flood / jitter) against a live VM. Hard per-op timeouts make hangs failures; end-of-run oracles check exactly-once nonces, no leaked processes, no leaked state. `make fuzz SEEDS=n OPS=m` replays any seed; one quick seed runs in the normal suite; `weekly.yml` gains a `Lifecycle-Fuzz-arm64` job running 20 seeds x 25 ops. The nextest override pins `retries = 0` for the chaos test: a caught race must fail the run, not vanish on retry (the exact hedge this suite exists to eliminate).

## Bugs this suite caught (all now fixed on main)

The tiers were built against the racy code and reproduced each flake deterministically; the fixes merged first, so this PR lands green:

- **Exec orphaned across a snapshot pause** — hidden `test_custom_command_bridged` flake (exec accepted in the ~40ms accept-to-pause window of startup-snapshot creation was silently orphaned by the vsock reset; the retry passed because the failed run had populated the snapshot cache). Fixed by the three-phase request/ACK/GO handshake + absolute ACK bound + vsock-epoch orphan detection (#710). Pinned by Tier 0 and `exec_resend_across_agent_stall` / `go_floor_across_ack_stall`.
- **Healthy published while vCPUs paused** — clients (port-forward curl, exec, `fcvm ls` pollers) could observe Healthy during the startup-snapshot pause. Fixed by deferring Healthy until the pause/resume cycle acks (#711). Pinned by `healthy_means_live_dataplane` and the chaos harness's Healthy⇒live chained curl.
- **Serial wedge from a mid-UART-TX snapshot** — the arm64 trio (`serial_console_after_restore`, `exec_rootless`, `exec_parallel_tty_stress`): the pre-start snapshot captured mid-UART-TX left a dead serial on every restore, and fc-agent console writes then wedged the exec accept loop after ~4KB. Fixed by the console quiesce-before-cache-ready + wedge-proof console writer (#712; fork-level UART restore fix tracked in #708). Pinned by `quiesce_survives_long_prepause`.
- Related groundwork this suite depends on: the CH ACPI shutdown contract (#706) which unblocked deterministic teardown oracles on x86.

## Rebase notes (hook re-siting onto the merged fixes)

The branch predates the final form of the fix PRs, so the hooks were re-sited into the merged code rather than kept verbatim: `snapshot.post_save_pre_resume` now holds AFTER #710's `record_vsock_reset_for_snapshot` epoch bump (preserving its bump-before-resume invariant), `health.pre_persist_healthy` sits after #711's dropped-ack "discard observation" gate, `cache_ready.pre_send` after #712's quiesce-or-abort gate, and the Tier 0 client harness now drives the post-#710 signature (`connect_and_start_exec(sock, req, vm_id)` → `(stream, SnapshotOrphanGuard)`, responses via `EpochGuardedReader`).

## Test evidence (aarch64, this branch = main + this commit)

```
make lint                                    # clean: cargo fmt --check (incl. failpoint),
                                             # clippy --all-targets -D warnings, cargo deny
make build                                   # clean (fcvm + fc-agent musl)

make test-unit
  Summary [  50.769s] 379 tests run: 379 passed, 0 skipped
  (12 failpoint crate tests, 21 Tier 0 protocol-fuzz tests,
   test_snapshot_key_golden + test_snapshot_key_changes_with_guest_failpoint included)

make test-root FILTER=lifecycle_interleave STREAM=1
  PASS [  29.490s] test_lifecycle_interleave_exec_resend_across_agent_stall
  PASS [  36.840s] test_lifecycle_interleave_go_floor_across_ack_stall
  PASS [  28.343s] test_lifecycle_interleave_healthy_means_live_dataplane
  PASS [  15.606s] test_lifecycle_interleave_no_leaks_after_interleaved_teardown
  PASS [  24.843s] test_lifecycle_interleave_quiesce_survives_long_prepause
  Summary [ 135.125s] 5 tests run: 5 passed, 657 skipped

make fuzz SEEDS=1 OPS=10                     # chaos smoke, retries=0 profile
  FUZZ ===== seed 1 starting (10 ops) =====
  FUZZ seed=1 nonce oracle OK (2 nonces, each exactly once)
  FUZZ seed=1 process-leak oracle OK
  FUZZ seed=1 state-file oracle OK
  FUZZ ===== seed 1 PASSED in 18.1s =====
  Summary [  18.114s] 1 test run: 1 passed, 661 skipped
```

## Summary by CodeRabbit

- **New Features**
  - Added configurable lifecycle testing controls for VM startup, execution, snapshots, health checks, and restoration.
  - Added seeded fuzz testing with configurable seeds and operation counts.
  - Added failure logs and replayable test traces for easier investigation.

- **Bug Fixes**
  - Improved handling of interrupted execution handshakes, retries, timeouts, and process cleanup.
  - Prevented stale VM state, orphaned processes, and duplicate command execution during lifecycle interruptions.

- **Tests**
  - Expanded coverage for snapshot, restore, health, shutdown, and execution interleavings.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -97,6 +97,21 @@ max-threads = "num-cpus"
 # stress_100 variant, each burning 100-150s before a clean retry). Give the
 # bench the entire runner: ~160s of exclusivity is cheaper in wall time than
 # the try-1 burns plus retries it causes.
+# Seeded lifecycle chaos fuzz (make fuzz, tests/test_fuzz_chaos.rs): ONE test
+# loops over every requested seed, each booting a VM and running a full op
+# schedule (~2-4 min/seed). The weekly CI run uses 20 seeds x 25 ops, so the
+# single test needs a multi-hour ceiling; the default run (1 seed, 10 ops)
+# finishes in ~2-3 min. Listed before the vm-tests override so it wins.
+[[profile.default.overrides]]
+filter = "package(fcvm) & test(=test_fuzz_lifecycle_chaos)"
+test-group = "vm-tests"
+slow-timeout = { period = "7200s", terminate-after = 1 }
+# NEVER retry the chaos fuzzer: its interleavings are seed-scheduled but not
+# cycle-deterministic, so a wholesale retry would almost always pass and turn a
+# caught race into a green job with only a "flaky" footnote — exactly the hedge
+# this suite exists to eliminate. A chaos failure must fail the run.
+retries = 0
+
 [[profile.default.overrides]]
 filter = "package(fcvm) & test(/8000_concurrent/)"
 test-group = "stress-tests"
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -31,6 +31,101 @@ jobs:
           path: /tmp/fcvm-container-target/criterion/
           if-no-files-found: warn
 
+  # Seeded lifecycle chaos fuzz (tests/test_fuzz_chaos.rs): 20 seeds x 25 ops
+  # against real rootless VMs. Schedule-deterministic per seed — a failure names
+  # the seed and is replayed locally with `make fuzz SEEDS=<seed>, OPS=25`.
+  fuzz:
+    name: Lifecycle-Fuzz-arm64
+    runs-on: [self-hosted, Linux, ARM64]
+    steps:
+      # Fix ownership of root-owned files from previous test runs (sudo test runner)
+      - name: Fix workspace permissions (pre-checkout)
+        run: |
+          sudo chown -R $USER:$USER ${{ github.workspace }}/fcvm 2>/dev/null || true
+          sudo chmod -R u+w ~/.cargo/advisory-db* 2>/dev/null || true
+          sudo chown -R $USER:$USER ~/.cargo/advisory-db* 2>/dev/null || true
+      - uses: actions/checkout@v6
+        with:
+          path: fcvm
+      - uses: ./fcvm/.github/actions/checkout-deps
+      - name: Install Rust
+        run: |
+          curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused --location --silent --show-error --fail https://sh.rustup.rs | sh -s -- --default-toolchain none -y
+          echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+          # rust-toolchain.toml handles version + components; just trigger install
+          rustup show active-toolchain || true
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y fuse3 libfuse3-dev libclang-dev clang musl-tools \
+            iproute2 iptables passt dnsmasq qemu-utils e2fsprogs parted patch \
+            podman skopeo busybox-static cpio zstd autoconf automake libtool cmake \
+            libseccomp-dev
+      - name: Build passt from source
+        working-directory: fcvm
+        run: ./scripts/build-passt.sh
+      - name: Install Firecracker
+        run: |
+          ARCH=$(uname -m)
+          FC_VERSION="1.14.0"
+          curl -L -o /tmp/firecracker.tgz \
+            "https://github.com/firecracker-microvm/firecracker/releases/download/v${FC_VERSION}/firecracker-v${FC_VERSION}-${ARCH}.tgz"
+          sudo tar -xzf /tmp/firecracker.tgz -C /usr/local/bin --strip-components=1 \
+            "release-v${FC_VERSION}-${ARCH}/firecracker-v${FC_VERSION}-${ARCH}" \
+            "release-v${FC_VERSION}-${ARCH}/jailer-v${FC_VERSION}-${ARCH}"
+          sudo mv "/usr/local/bin/firecracker-v${FC_VERSION}-${ARCH}" /usr/local/bin/firecracker
+          sudo mv "/usr/local/bin/jailer-v${FC_VERSION}-${ARCH}" /usr/local/bin/jailer
+      - name: Install cargo tools
+        run: |
+          which cargo-nextest || cargo install cargo-nextest@0.9.115 --locked
+          sudo ln -sf $HOME/.cargo/bin/cargo /usr/local/bin/
+          sudo ln -sf $HOME/.cargo/bin/rustc /usr/local/bin/
+          sudo ln -sf $HOME/.cargo/bin/cargo-nextest /usr/local/bin/
+      - name: Setup KVM and networking
+        run: |
+          sudo chmod 666 /dev/kvm
+          sudo mkdir -p /var/run/netns
+          if [ ! -e /dev/userfaultfd ]; then
+            sudo mknod /dev/userfaultfd c 10 126
+          fi
+          sudo chmod 666 /dev/userfaultfd
+          sudo sysctl -w vm.unprivileged_userfaultfd=1
+          sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0 2>/dev/null || true
+          sudo sysctl -w net.ipv4.conf.all.forwarding=1
+          sudo sysctl -w net.ipv4.conf.default.forwarding=1
+          sudo sysctl -w net.ipv6.conf.all.forwarding=1
+          sudo sysctl -w net.ipv6.conf.default.forwarding=1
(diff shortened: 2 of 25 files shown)


ejc3:
## Fixed the SnapshotDisabled lane failures (root cause, no skips)

**Failure** (Host-Root-arm64/x64-SnapshotDisabled, run 31164334105): three tests failed deterministically —

- `test_lifecycle_interleave_healthy_means_live_dataplane` — panicked at the non-vacuity assert: *"the startup snapshot must be created THIS run (cold boot) or the gate is never exercised"*
- `test_lifecycle_interleave_quiesce_survives_long_prepause` — timed out waiting for `Pre-start snapshot created successfully`
- `test_lifecycle_interleave_no_leaks_after_interleaved_teardown` — timed out (300s) waiting for `FAILPOINT snapshot.pre_pause reached`

Every failing VM's log shows `Snapshot disabled via FCVM_NO_SNAPSHOT environment variable`: the spawn helpers inherit the full test environment, so the lane's `FCVM_NO_SNAPSHOT=1` leaked into VMs whose test **subject** is the snapshot lifecycle — the exact path they assert was turned off underneath them.

**Fix**: the SnapshotDisabled lane exists to cover the no-snapshot path of *ordinary* tests. A test that pins the snapshot lifecycle must **opt into snapshots explicitly** — then it tests what it claims in every lane. Added `common::spawn_fcvm_snapshots_enabled_with_env_and_log_path` (removes `FCVM_NO_SNAPSHOT` from the child env via `Command::env_remove`, over a general `env_remove` inner) and switched the three tests' VM spawns to it. No `#[ignore]`, no env-conditional early returns, no filters — this replaces closed auto-fix PR #720, whose diagnosis was right but whose skip was the wrong fix.

The other two interleave tests are untouched: `exec_resend_across_agent_stall` drives a *manual* `fcvm snapshot create` (not gated by `FCVM_NO_SNAPSHOT`) and `go_floor_across_ack_stall` has no snapshot involvement — both already passed in the lane, and still exercise the lane's inherited env.

**Test evidence** (all 5 interleave tests, both ways, locally):

Run 1 — `make test-root FILTER=lifecycle_interleave` (snapshots enabled):
```
Summary [36.731s] 5 tests run: 5 passed, 657 skipped
```

Run 2 — `FCVM_NO_SNAPSHOT=1 make test-root FILTER=lifecycle_interleave` (simulating the failing lane):
```
PASS [15.724s] test_lifecycle_interleave_no_leaks_after_interleaved_teardown   (was: 300s timeout)
PASS [17.960s] test_lifecycle_interleave_exec_resend_across_agent_stall
PASS [24.740s] test_lifecycle_interleave_quiesce_survives_long_prepause       (was: 66s timeout)
PASS [25.014s] test_lifecycle_interleave_go_floor_across_ack_stall
PASS [27.782s] test_lifecycle_interleave_healthy_means_live_dataplane         (was: assert panic)
Summary [27.785s] 5 tests run: 5 passed, 657 skipped
```
Run-2 VM logs confirm the mechanism: spawn line records `env_removed=["FCVM_NO_SNAPSHOT"]`, no `Snapshot disabled via FCVM_NO_SNAPSHOT` line, and `Creating pre-start snapshot` fires.

`make lint` (fmt, clippy `-D warnings`, audit, deny): pass.

ejc3:
Merging with the [MEDIUM] docs finding deferred deliberately: AGENTS.md/DESIGN.md failpoint documentation lands with the failpoint-coverage tranche (tracked task: hold-points for the new event-driven signals), which touches the same sections — one docs change instead of two CI cycles.

PR #719, opened 2026-08-07
Cut clone hot-path latency in half by removing per-launch execs and network RTTs

ejc3:
Halve `fcvm snapshot run` spawn->exec-ready latency by taking three per-launch costs off the clone hot path: **p50 679 ms -> 331 ms (-51%), stdev 10.9 ms -> 1.7 ms, execve() per clone 35 -> 19** (Graviton3, alpine, rootless, 2 rounds x 10 clones, interleaved A/B).

## The Problem

Every `fcvm snapshot run` paid for work whose answer almost never changes:

1. **`git ls-remote` to GitHub on every launch** (~340 ms — half the total). `get_profile_firecracker_path()` re-resolved the firecracker fork branch remotely per call, on a path that runs per clone.
2. **Repeated `--version` / `ldd --version` probes.** `find_firecracker()` runs more than once per launch, and each probe fork+execs a multi-megabyte binary whose version only changes when the binary does.
3. **~10 per-command `ip` fork+execs** configuring the rootless namespace, plus a second `nsenter`+`ip` just to verify the TAP device exists.

## The Solution

**1. On-disk resolved-profile cache** (`src/setup/kernel.rs`)
- Last remote resolution persisted at `assets_dir/firecracker/<profile>.resolved.json`; reused when repo/branch match, the TTL (`FCVM_FIRECRACKER_RESOLVE_TTL_SECS`, default 3600; `0` = always re-query) has not expired, and the binary still exists. Corrupt/unreadable cache is a miss, never an error.
- Only written when the resolved binary exists, so a not-yet-built profile keeps re-resolving (and failing loudly) instead of caching a dead path.
- `ensure_profile_firecracker()` — the `fcvm setup` path — always performs the remote resolution and rewrites the cache at **all three** of its success returns, so a rebuilt fork takes effect immediately, not after the TTL.
- Writes are unique-temp (uuid) + atomic rename, **never flock** — `assets_dir` can be shared across nesting levels over fuse-pipe, where flock does not cross the VM boundary (the #677 cross-VM cache lesson).
- The interleaved nocache control arm (`FCVM_FIRECRACKER_RESOLVE_TTL_SECS=0`) isolates this optimization at **-340.6 ms**.

**2. Version-probe memoisation** (`src/version_cache.rs`, new)
- `<binary> --version` output keyed by binary identity (path + mtime + size): process-local map + `assets_dir/version-cache/<key>.json`, so the short-lived per-clone fcvm process skips the exec too.
- Only successful probes are cached — `find_cloud_hypervisor()`'s fallback depends on failing binaries continuing to fail. A rebuilt binary changes identity and is re-probed. Cache IO problems degrade to running the binary, never fail the probe.
- `libc_version_tag()` and `detect_host_ipv6()` memoised per process (`OnceLock`).

**3. `ip -batch` consolidation** (`src/network/pasta.rs`, `src/commands/podman/namespace.rs`, `src/commands/common.rs`)
- New `IpBatchScript` renders each setup phase's `ip` commands as one `ip -batch` heredoc: one `nsenter`+`bash`+`ip` per phase instead of a process per command.
- The TAP existence check is the **last batch line** — `ip -batch` aborts at the first failure, so reaching it proves every prior step applied — replacing the separate verification exec in both the fresh-VM path and the clone path.
- On failure, `ip -batch`'s `Command failed -:<line>` is mapped back to the step's description (`describe_failure`), so errors still name the failing step; unrecognised stderr passes through verbatim.

**New benchmark**: `bench/clone-latency.sh` + `bench/clone-latency-report.py` (`make bench-clone-latency LABEL=x N=10`) — times N `snapshot run` cycles from spawn until fc-agent's exec server is ready (event-driven `tail -F | grep -m1` on the debug log, no polling in the measured path) and prints a stage breakdown parsed from the `RUST_LOG=fcvm=debug` timeline. Results in `/tmp/fcvm-clone-latency-<label>/`, never committed.

**Docs**: PERFORMANCE.md stage table + optimization notes; README.md documents `FCVM_FIRECRACKER_RESOLVE_TTL_SECS` and the resolution-cache contract.

## Stage breakdown (from PERFORMANCE.md, one 10-clone round)

| Stage | Before | After |
|-------|--------|-------|
| firecracker resolution (`git ls-remote`) | 337 ms | 0.4 ms |
| netns + pasta setup | 101 ms | 95 ms |
| snapshot load + resume | 6 ms | 6 ms |
| resume -> exec-server-ready | 222 ms | 221 ms |
| **total (p50)** | **673 ms** | **331 ms** |
| `execve()` per clone | 35 | 19 |

The remaining `resume -> exec-server-ready` (~221 ms) is guest-side fc-agent restore work — the next target.

## Test Results

```
$ make lint
cargo fmt -p fcvm -p fuse-pipe -p fc-agent --check   # clean
cargo clippy --all-targets -- -D warnings            # clean

$ make test-unit
Summary [  50.478s] 364 tests run: 364 passed, 0 skipped
# includes the 19 new tests: 8 kernel resolution-cache, 5 pasta ip-batch,
# 5 version_cache, TTL-env parser

Summary [1221.965s] 113 tests run: 113 passed, 528 skipped

$ make test-root FILTER=sanity
Summary [  18.191s] 5 tests run: 5 passed, 636 skipped
```

## Post-rebase bench (this branch rebased onto d85675da, i.e. current main with #706/#710/#711/#712)

```
$ make bench-clone-latency LABEL=post-rebase N=10
  p50   332.0 ms   (min 330.0, max 337.0, p90 334.0, mean 332.4)

stage breakdown (median of per-clone deltas, ms)
  resolve-fc              0.4      # ls-remote cache holds after rebase
  listeners+state         2.0
  netns+pasta            95.7
  snapshot-load           6.5
  resume->ready         221.6
```

## Summary by CodeRabbit

* **Performance**
  * Reduced documented median VM clone startup latency from 673 ms to 331 ms.
  * Improved repeated launches with cached binary and version lookups and faster network setup.
  * Improved restore reliability with unique restore identifiers.

* **New Features**
  * Added a configurable clone-latency benchmark with detailed stage reports and exportable samples.

* **Documentation**
  * Documented benchmark usage, Firecracker resolution caching, cache expiration settings, and cache behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ CONTAINER_RUN := $(CONTAINER_RUN_BASE) --ulimit nproc=65536:65536 --pids-limit=6
 	container-build container-test container-test-unit container-test-fast container-test-all container-test-fc-mock \
 	container-setup-fcvm container-shell container-clean container-bench \
 	setup-btrfs setup-default setup-fcvm setup-pjdfstest setup-hugepages bench bench-vm bench-hugepages bench-hugepages-test \
-	bench-container-import bench-chromium \
+	bench-container-import bench-chromium bench-clone-latency \
 	lint fmt update-dependency ssh test-serve-sdk
 
 all: build
@@ -234,6 +234,7 @@ help:
 	@echo "  bench-hugepages-test  Run hugepages benchmark (2GB VM, 256MB dirty)"
 	@echo "  bench-container-import  Compare podman load vs direct image mount"
 	@echo "  bench-chromium     Chromium shared-nothing clone bench (egress x memory matrix)"
+	@echo "  bench-clone-latency  Clone spawn->exec-ready latency (LABEL=, N=)"
 	@echo ""
 	@echo "Other:"
 	@echo "  lint               Run linting (auto-installs tools if needed)"
@@ -634,6 +635,15 @@ bench-hugepages-test: build setup-default
 bench-container-import: build setup-default
 	@echo "==> Running container import benchmark..."
 	$(CARGO) bench --bench container_import
+
+# Clone hot-path latency: `fcvm snapshot run` spawn -> exec-server-ready, N times.
+# Event-driven readiness (waits on the debug log's marker), plus a stage breakdown
+# parsed from the RUST_LOG=fcvm=debug timeline.
+# Results in /tmp/fcvm-clone-latency-$(LABEL)/ (never committed).
+#   make bench-clone-latency LABEL=before N=10
+bench-clone-latency: build setup-default
+	@echo "==> Running clone latency benchmark..."
+	bench/clone-latency.sh $(or $(LABEL),run) $(or $(N),10)
 
 # Container benchmark target (used by nightly CI)
 # Uses CONTAINER_RUN_BASE (no --ulimit nproc) to avoid EPERM on GHA ubuntu-latest
--- a/PERFORMANCE.md
+++ b/PERFORMANCE.md
@@ -248,6 +248,49 @@ All 10 clones launched simultaneously:
 
 Core restore is ~10ms regardless of sequential or parallel. Single memory server handles 10 concurrent clones. Bottleneck is network teardown and state cleanup under contention.
 
+### Clone Spawn → Exec-Ready Latency
+
+`bench/clone-latency.sh <label> <N>` boots a minimal alpine VM, snapshots it,
+serves it, and times N `fcvm snapshot run` cycles from process spawn until
+fc-agent's exec server is reachable (event-driven: it waits on the debug log's
+readiness marker, no polling). It prints p50/min/max plus a stage breakdown
+parsed from the `RUST_LOG=fcvm=debug` timeline; results land in
+`/tmp/fcvm-clone-latency-<label>/` and are not committed.
+
+```bash
+make build
+bench/clone-latency.sh mylabel 10
+```
+
+Measured on Graviton3, alpine, rootless networking, 2 rounds × 10 clones:
+
+| Stage | Before | After |
+|-------|--------|-------|
+| firecracker resolution (`git ls-remote`) | 337 ms | 0.4 ms |
+| netns + pasta setup | 101 ms | 95 ms |
+| snapshot load + resume | 6 ms | 6 ms |
+| resume → exec-server-ready | 222 ms | 221 ms |
+| **total (p50)** | **673 ms** | **331 ms** |
+| `execve()` per clone | 35 | 19 |
+
+Three hot-path costs were removed, all of them work that does not belong in a
+per-launch path:
+
+1. **`git ls-remote` to GitHub on every launch** (~337 ms, half the total). The
+   fork-branch resolution is now cached under `assets_dir/firecracker/` with a
+   TTL (`FCVM_FIRECRACKER_RESOLVE_TTL_SECS`, default 1h); `fcvm setup` always
+   re-queries and rewrites it, so a rebuilt fork is picked up immediately.
+2. **Repeated `--version` / `ldd --version` probes.** Cached per binary identity
+   (path + mtime + size), in-process and on disk, so a changed binary still
+   re-checks. Removes 3 execs per clone.
+3. **Per-command `ip` execs in the rootless namespace.** Each setup phase now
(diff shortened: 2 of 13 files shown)


ejc3:
## Root cause of the SnapshotEnabled-arm64 failures: wall-clock restore-epoch collision (deaf clones)

**TL;DR:** this PR made clone restores fast enough that a *snapshot of a restored VM* could be restored again **within the same wall-clock second**. The restore-epoch was `SystemTime::now().as_secs()`, so the second restore handed the guest an epoch **identical** to the one already recorded in the snapshot's memory — fc-agent's watcher (pure string-inequality compare) saw no change, never ran `handle_clone_restore`, never reconnected exec/egress/output vsocks, and the clone never became healthy. Fixed in 4bf3f137 by minting the epoch as a per-restore UUID.

### Smoking gun

Reproduced deterministically on an idle 64-core host with `test_snapshot_restore_survives_ancestor_dir_cleanup` (2/2 FAIL at 120.8s — the same test/signature that failed the Host-arm64 lane in run 31165538739):

```
[sib608-a] 12:20:06.042 signaled fc-agent via MMDS restore_epoch=1786105206
[sib608-a] 12:20:06.171 [fc-agent] restore complete (epoch=1786105206)
    ... snapshot T1 taken from A captures last_epoch=1786105206 in guest memory ...
[sib608-b] 12:20:06.785 signaled fc-agent via MMDS restore_epoch=1786105206   <-- IDENTICAL
    ... nothing ever again from B's guest: vcpus idle in WFI, no reconnect, never healthy
```

### Why it nuked run 2 (and only run 2) of Host-Root-arm64-SnapshotEnabled

Run 2 runs with **warm snapshots**, so every test baseline is itself a restore (fc-agent's watcher holds `last_epoch = Some(<second of the baseline's restore>)`). Every snapshot-then-clone test then creates a snapshot from that baseline and restores it — on this branch, inside the same second → deaf clone → full 60–180s health timeout with a live firecracker idling. Run 2's summary shows the fallout: **28 failed + 19 timed out** (`test_snapshot_clone` ×9 at exactly ~120.1s, `hugepage_snapshot_clone`, reboot/diff chains, pjdfstest ×15…). The job's own `pgrep` dumps show firecracker count climbing **88 → 123** as deaf VMs stacked up; the resulting contention (120–180s waits on the loopback-ip flock, `cp --reflink` stalled for minutes) then timed out everything still queued — which is why the tail-end `startup_snapshot` tests died at uniform ~310s during *boot*, not in their own logic. Run 1 (cold boots: watcher state `None`, so the first epoch always triggers) was green on the same host minutes earlier, and x64-SnapshotEnabled squeaked by on timing.

### Fix (4bf3f137)

- `new_restore_epoch()` returns `Uuid::new_v4()` — unique per restore. fc-agent already treats the epoch as an opaque string (its own unit test parses `"abc123"`), so **no guest-side change**.
- Both producers switched: the Firecracker MMDS PUT (`commands/common.rs`) and the Cloud Hypervisor boot-plan listener (`commands/snapshot.rs`).
- Regression guard: `restore_epochs_differ_for_back_to_back_restores` fails if the epoch ever regresses to second granularity.

### Verification

- `make lint` clean; new unit test passes.
- `make test-root FILTER=ancestor_dir_cleanup`: **PASS in 17.4s** (was 2/2 FAIL at 120.8s before the fix). Logs now show three distinct UUID epochs with `restore complete` for every link of the A→B→C restore chain, 0.8s apart.
- `make test-root` over `ancestor_dir_cleanup|startup_snapshot|snapshot_run_direct|snapshot_run_exec|hugepage_snapshot_clone`: **14/14 passed**, including `test_startup_snapshot_on_restored_vm` and `test_startup_snapshot_priority`.

Note: the attempt-2 rerun that was queued still runs the old commit and may fail the same way; the push supersedes it.

ejc3:
Merging on user direction. Confidence basis: interleaved A/B (679→331ms p50, stdev 10.9→1.7), lint clean, snapshot suite 113/113 including the UUID restore-epoch regression test, sanity 5/5. This branch also carries the deaf-clone fix — sub-second restores exposed that wall-clock-second restore epochs collide, leaving clones permanently deaf; root-caused with a deterministic repro and fixed with per-restore UUIDs. Its only CI failure was on the deregistered disk-full runner.

PR #722, opened 2026-08-07
Fix ~1s IPv6 DAD stall on routed clone restore

ejc3:
Eliminates the ~1s IPv6 egress blackout every routed-mode clone restore hit before its first outbound connection, by keeping Duplicate Address Detection off the routed path's nexthop resolution.

## The Problem

A restored routed clone is healthy in ~700ms, but its first egress connection stalled another 0.6-1.6s. Measured with a 10ms event-driven in-guest probe: routed clone net-ready p50 **1,002.5ms** vs **0.7ms** after the fix (n=8 each); rootless and bridged were already fast and are unaffected.

### Root cause: IPv6 DAD on both ends of nexthop resolution

Routed guest traffic leaves via the namespace's IPv6 default route, whose nexthop is the **host veth's link-local**. Resolving that nexthop needs a Neighbor Solicitation sourced from a valid link-local of the namespace's own (**br0**). DAD made both ends `tentative`:

1. **Host veth**: the moment both veth ends came up, the kernel autogenerated the EUI-64 link-local itself — tentative, mid-DAD. fcvm's own `nodad` add of the same address then failed EEXIST, so the `nodad` flag never landed. A tentative address does not answer NS.
2. **Namespace br0**: its link-local also sat in DAD, so it could not *source* the NS to resolve the nexthop either.

### Evidence the stall is DAD, not something else

| Observation | DAD prediction | Fit |
|---|---|---|
| Stall duration 0.6-1.6s, ~1.0s spread | rand(0..router_solicitation_delay=**1s**) kick + dad_transmits(1) × retrans_time_ms(**1000**) | uniform ~1s spread matches the random kick window |
| `assigning link-local to host veth failed ... File exists` warned in **24/24** benchmark logs | kernel autogenerates the same EUI-64 address first, so fcvm's `nodad` add must fail EEXIST every time | 24/24 |
| Fresh boots never showed it; only restores | DAD runs in the shadow of a multi-second cold boot, but lands on the critical path of a ~700ms restore | matches |
| Rootless/bridged unaffected | only routed resolves an IPv6 link-local nexthop over a veth | matches |
| p50 1,002.5ms → 0.7ms with the fix (n=8 each) | removing DAD removes the whole stall | matches |

## The Solution (100% host-side, `src/network/routed.rs`)

- **Step 1b (new)**: `net.ipv6.conf.{all,default}.accept_dad=0` inside the VM's network namespace, before any interface exists there. Both sysctls are namespace-local, so this is scoped entirely to the VM's own private point-to-point segment — nothing on the host or other VMs is touched.
- **Step 2b (new)**: `net.ipv6.conf.<veth>.addr_gen_mode=1` on the host veth **before link-up** (`create_veth_pair` leaves both ends down, so the window is race-free) — the kernel never claims the link-local first, and step 8's explicit `nodad` add is the one that takes effect, instantly valid. Plus `keep_addr_on_down=1` as self-healing insurance: with autogeneration off, the kernel's default flush-on-link-down would otherwise lose the manually-installed nexthop address for good on a veth bounce.
- **Step 8 (rewritten)**: `assign_link_local_nodad()` installs the EUI-64 link-local with `nodad`. If the add fails **and the interface really does already hold the address** (older kernel rejecting `addr_gen_mode`, or a leftover on a recycled interface name), it deletes and re-adds — `ip addr replace/change` set the NODAD flag but do **not** cancel in-flight DAD, so del+add is the only repair that yields a non-tentative address. The already-present case is decided from interface state (`ip -j -6 addr show`, addresses compared parsed), not iproute2 error text; any other add failure propagates without deleting anything.

### ⚠️ Behavior change: warn-and-continue → hard error

Step 8 previously logged a warning and kept going when the link-local add failed. It now **fails VM setup** if the nexthop address cannot be made usable. This is a deliberate new failure class: a routed VM whose IPv6 default-route nexthop is unusable has silently broken egress, and failing fast beats coming up broken.

## Regression Test

`test_egress_clone_routed_no_dad_stall` (tests/test_egress.rs) restores a routed clone and asserts:

1. The host veth holds **exactly** the EUI-64 link-local (computed independently from the MAC in the test), its `inet6` line carries `nodad`, and it is not `tentative` — a kernel-autogenerated address can never satisfy this, so a regression fails **every** run, not just when the random DAD delay is long. Verified: with the fix reverted, this assertion failed **2/2**; with the fix it passes.
2. Nothing in the clone's namespace is tentative (br0 must be able to source the NS).
3. In-guest egress to a host IPv6 listener works (functional check; each connect attempt capped at 50ms so a black-holed SYN fails the test instead of hanging it).

There is deliberately **no timing assertion**: by the time the harness can exec into the clone (~0.5-1s of health polling and state lookups after the veth came up) the DAD window has already expired, so any wall-clock bound measured from there would pass with the bug present.

`tests/common/mod.rs`: `get_loopback_ip()` generalized into `get_network_field()` so the test can read `host_veth`/`namespace_name`.

## Test Results

```
$ make lint
advisories ok, bans ok, licenses ok, sources ok   # clippy + fmt clean

$ make test-root FILTER=egress_clone_routed STREAM=1
PASS [  33.584s] (1/2) fcvm::test_egress test_egress_clone_routed
PASS [  17.840s] (2/2) fcvm::test_egress test_egress_clone_routed_no_dad_stall
Summary [  51.427s] 2 tests run: 2 passed, 622 skipped

$ make test-root FILTER=egress_fresh_routed STREAM=1
PASS [   6.469s] (1/1) fcvm::test_egress test_egress_fresh_routed
Summary [   6.472s] 1 test run: 1 passed, 623 skipped

$ make test-root FILTER=egress_fresh_rootless STREAM=1     # rootless regression spot-check
PASS [  21.612s] (1/1) fcvm::test_egress test_egress_fresh_rootless
Summary [  21.614s] 1 test run: 1 passed, 623 skipped
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routed networking for IPv6 by preventing duplicate-address detection stalls.
  * Preserved manually assigned IPv6 addresses when interfaces are activated or restored.
  * Ensured host link-local addresses are configured reliably without masking unrelated errors.

* **Tests**
  * Added integration coverage for IPv6 connectivity in restored routed VM clones.
  * Added validation that IPv6 addresses are not left in a tentative state.


ejc3:
Merging on user direction. Confidence basis: 4/4 local tests read and passing (egress_clone_routed 2/2 including the new no-DAD-stall regression test, egress_fresh_routed 1/1, rootless spot-check 1/1), lint clean, adversarial review with both SHOULD-FIX findings applied (EEXIST-only repair path, honest test assertions). The only CI failure was Host-Root-arm64-SnapshotDisabled on the deregistered disk-full runner. BEHAVIOR CHANGE called out in the PR body: step 8 is now a hard error rather than warn-and-continue.

PR #729, opened 2026-08-07
Add disk-capacity preflight + hourly runner guard so a full runner can't poison jobs

ejc3:
Makes a disk-full self-hosted runner a loud, self-healing condition instead of a silent job-poisoner.

## The Problem

On 2026-08-06 `runner-i-0ea78b39da373fcca` filled its disk and then failed **every job it picked up** during GitHub's `Set up job` phase:

```
No space left on device: /opt/actions-runner/_diag/...
```

Two of the poisoned jobs were on the dependabot merge train. The runner had to be manually deregistered.

The critical detail for the fix: **the runner died before any job step ran.** A job-level preflight step could not have saved it — there was no step. So a single-layer fix is not a fix.

### (a) `ci.yml` preflight step

New `Disk preflight` step in the three self-hosted jobs (`host`, `host-root`, `container`), placed right after checkout and before the dependency clones and builds that consume the space. GitHub-hosted jobs (`lint`, `packaging`, `fc-mock`) are untouched — they're ephemeral.

Below `MIN_FREE_GB` (default 15) it cleans in order of increasing cost-to-recreate, stopping as soon as space is recovered:

1. `podman system prune -af --volumes` (rootless + root stores)
2. stale `/tmp/fcvm-test-logs` (> `LOG_AGE_DAYS`, default 1)
3. idle cargo/nextest target dirs — runner `_work` checkouts and `/mnt/fcvm-btrfs/cargo-target*` (> `TARGET_AGE_DAYS`, default 3)
4. content-addressed btrfs caches — `image-cache`, snapshots (incl. the root/container data dirs), kernel/rootfs/initrd/firecracker assets, core dumps (> `CACHE_AGE_DAYS`, default 4)

Still below `HARD_FLOOR_GB` (default 5) afterwards → **fails the job loudly** with a `::error` annotation naming the runner, so it gets attention instead of producing mystery failures downstream. `df` is echoed before and after; every deleted entry is logged with its byte size.

This step needs **no secrets** and adds no `${{ github.event.* }}` interpolation into `run:`.

### (b) Runner-level guard — the layer that actually covers the outage

`scripts/runner-disk-guard.{service,timer}`: an hourly systemd timer running the **same script** with `--quarantine`, installed into the runner AMI (`build-ami.sh`) and by the bootstrap (`setup-runner.sh`). Running out-of-band of any job, it cleans before the box ever reaches setup-fails territory; if the hard floor still can't be met it stops and disables the `actions.runner.*` unit so the runner goes **offline** and the autoscaler replaces it, rather than accepting and killing jobs.

The AMI/bootstrap are defined in this repo, so no `ejc3/aws` change is needed.

**`build-ami.sh` cache-hash fix (required, not incidental):** `compute_hash()` didn't hash the user-data or the installed files, so the cached guard-less AMI would have been reused and layer (b) would never have shipped. It now also hashes the guard script/units and `create_user_data`. Verified: old logic returns an **unchanged** hash on this branch (`005d73a75a65`), new logic returns `ed4dddefe223`.

## Two subtleties the local dry run exposed

Both are fixed and commented in-script, because both would have silently degraded the guard:

- **The idle probe must use `-type f`.** Reading a directory updates that directory's atime — so a probe based on directory atime refreshes the very timestamps it reads, and after one preflight run *nothing would ever look idle again*. File atimes are untouched by directory traversal, making them the only self-consistent signal.
- **`[[ -d ]]` lies here.** The root/container data dirs are root-owned and not traversable by the CI user, so a plain test reported "missing" and **silently skipped the largest cache on the box** (814 snapshot dirs). Existence is now tested through `sudo`.

## Safety of the cache pruning

These are shared content-addressed assets, so the age gate requires **no read AND no write** for a generous margin. Anything the current checkout's SHAs reference is re-read every CI run (relatime keeps atime fresh) and is therefore never a candidate; a pruned entry is regenerated on the next cache miss. `image-cache` deletions additionally take the same per-digest `flock` the image builders use, and skip entries whose lock is held. `state/`, `vm-disks/` and `*.lock` files are never touched. Failures are never silenced — the only tolerated absences are "target legitimately may not exist", each commented.

## Test Results

Run for real on a dev box with podman + `/mnt/fcvm-btrfs`.

**Lint:**
```
$ shellcheck scripts/runner-disk-preflight.sh && bash -n scripts/runner-disk-preflight.sh
SHELLCHECK-CLEAN
# build-ami.sh: 4 findings before and 4 after -- no new findings
$ systemd-analyze verify runner-disk-guard.service runner-disk-guard.timer   # both clean
```

**Dry run (`DRY_RUN=1`, audits all stages regardless of thresholds):**
```
[disk-preflight] monitored filesystems: / /mnt/fcvm-btrfs
... 814 root/snapshots dirs scanned, all "keeping (active within 4d)"
[image-cache entries (idle >4d)]  0 entries, 0B would be freed
[idle cargo target dirs (>3d)]    0 entries, 0B would be freed
[idle kernels assets (>4d)]       0 entries, 0B would be freed
```
Zero candidates — I verified by hand this is correct rather than a broken probe: every entry is genuinely from the previous day (`mtime == atime`), so nothing stale was being masked, and nothing precious (current-SHA kernels/rootfs, live VM state/disks) appeared in the candidate list.

**Real run, thresholds set to force the podman-prune path:**
```
$ MIN_FREE_GB=99999 HARD_FLOOR_GB=1 ./scripts/runner-disk-preflight.sh    # exit 0
=== df -h BEFORE ===
/dev/root       290G  248G   42G  86% /
/dev/nvme1n1    3.5T  671G  2.9T  19% /mnt/fcvm-btrfs

=== stage_podman_prune ===        -> pruned 114 rootless + 37 root images

=== df -h AFTER ===
/dev/root       290G  248G   42G  86% /
/dev/nvme1n1    3.5T  668G  2.9T  19% /mnt/fcvm-btrfs
```
3.1 GiB reclaimed. btrfs frees extents asynchronously, so the bytes land in a later stage's `df` delta — noted in-script so nobody debugs a phantom attribution later.

**Hard-floor failure + quarantine:**
```
$ MIN_FREE_GB=99999 HARD_FLOOR_GB=99999 ./scripts/runner-disk-preflight.sh --quarantine
EXIT=1
[disk-preflight] FATAL: runner 'ip-10-0-1-49' still below 99999GiB free after cleanup
::error title=Runner disk full::Runner 'ip-10-0-1-49' is below 99999GiB free after cleanup - replace or manually clean this runner
[disk-preflight] quarantine: no actions.runner service unit on this host
```

**Default thresholds / arg handling:**
```
$ ./scripts/runner-disk-preflight.sh
[disk-preflight] OK: every monitored filesystem has >= 15GiB free; no cleanup needed
EXIT=0
$ ./scripts/runner-disk-preflight.sh --bogus    ->  EXIT=2
```

**Workflow validation** (no `actionlint`/`yq` on this box, so parsed with PyYAML):
```
host:      step[2] = Disk preflight ; prev=actions/checkout@v6 ; next=checkout-deps
host-root: step[2] = Disk preflight ; prev=actions/checkout@v6 ; next=checkout-deps
container: step[2] = Disk preflight ; prev=actions/checkout@v6 ; next=checkout-deps
GitHub-hosted jobs correctly untouched
```

## Merge note

Ideal merge-train candidate by nature, but it **touches `ci.yml`**, so it needs its own full matrix run anyway.

## Summary by CodeRabbit

* **New Features**
  * Added automated disk-capacity monitoring and cleanup for CI runners.
  * Added scheduled protection that runs shortly after startup and hourly thereafter.
  * Runners can be quarantined when disk space remains critically low.
* **Bug Fixes**
  * CI jobs now check available disk space before dependency checkout and builds.
  * AMI provisioning now includes the disk-protection components and reflects related changes in image caching.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,18 @@ jobs:
       - uses: actions/checkout@v6
         with:
           path: fcvm
+      # Free space before the dependency clones and builds that need it, and
+      # fail loudly (naming the runner) if the disk can't be recovered — a
+      # full runner otherwise produces mystery ENOSPC failures downstream.
+      # Runs as early as the repo's own script is available; it cannot rescue
+      # a runner already too full for "Set up job" itself (2026-08-06:
+      # runner-i-0ea78b39da373fcca died there on _diag/ and killed every job
+      # it picked up, including two on the dependabot merge train). That case
+      # is covered on the runner by scripts/runner-disk-guard.timer, which
+      # runs this same script hourly with --quarantine. Needs no secrets.
+      - name: Disk preflight
+        working-directory: fcvm
+        run: ./scripts/runner-disk-preflight.sh
       - uses: ./fcvm/.github/actions/checkout-deps
       - name: Install Rust
         run: |
@@ -463,6 +475,18 @@ jobs:
       - uses: actions/checkout@v6
         with:
           path: fcvm
+      # Free space before the dependency clones and builds that need it, and
+      # fail loudly (naming the runner) if the disk can't be recovered — a
+      # full runner otherwise produces mystery ENOSPC failures downstream.
+      # Runs as early as the repo's own script is available; it cannot rescue
+      # a runner already too full for "Set up job" itself (2026-08-06:
+      # runner-i-0ea78b39da373fcca died there on _diag/ and killed every job
+      # it picked up, including two on the dependabot merge train). That case
+      # is covered on the runner by scripts/runner-disk-guard.timer, which
+      # runs this same script hourly with --quarantine. Needs no secrets.
+      - name: Disk preflight
+        working-directory: fcvm
+        run: ./scripts/runner-disk-preflight.sh
       - uses: ./fcvm/.github/actions/checkout-deps
       - name: Install Rust
         run: |
@@ -649,6 +673,18 @@ jobs:
       - uses: actions/checkout@v6
         with:
           path: fcvm
+      # Free space before the dependency clones and builds that need it, and
+      # fail loudly (naming the runner) if the disk can't be recovered — a
+      # full runner otherwise produces mystery ENOSPC failures downstream.
+      # Runs as early as the repo's own script is available; it cannot rescue
+      # a runner already too full for "Set up job" itself (2026-08-06:
+      # runner-i-0ea78b39da373fcca died there on _diag/ and killed every job
+      # it picked up, including two on the dependabot merge train). That case
+      # is covered on the runner by scripts/runner-disk-guard.timer, which
+      # runs this same script hourly with --quarantine. Needs no secrets.
+      - name: Disk preflight
+        working-directory: fcvm
+        run: ./scripts/runner-disk-preflight.sh
       - uses: ./fcvm/.github/actions/checkout-deps
       - name: Setup KVM and rootless podman
         run: |
--- a/scripts/build-ami.sh
+++ b/scripts/build-ami.sh
@@ -17,6 +17,15 @@ compute_hash() {
     # Include the passt build inputs so a pin or patch change rebuilds the AMI
     # instead of reusing one whose baked-in pasta no longer matches CI.
     cat "$SCRIPT_DIR/build-passt.sh" "$SCRIPT_DIR"/passt-*.patch 2>/dev/null
+    # Include the disk guard's script and units: they are baked into the AMI,
+    # so a change to them must produce a new AMI rather than silently reusing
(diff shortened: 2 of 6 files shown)


ejc3:
Merging on user direction (confident + low blast radius). CI is currently blocked by a self-hosted runner-pool collapse (8→2 runners, disk-full failures), so remaining checks are queued rather than red — every failure on this PR was attributed to the deregistered disk-full runner, not to the change. Local verification stands in the PR body.

PR #730, opened 2026-08-07
Restore the PR_SET_PDEATHSIG teardown chain across the sudo privilege hop

ejc3:
Restore the PR_SET_PDEATHSIG teardown chain across the sudo privilege hop

`make test-root` keeps cargo-nextest unprivileged and elevates only the test
binary (CARGO_TARGET_*_RUNNER='sudo -E env PATH=...'), so the CI process tree is:

  nextest (uid 1000) -> sudo (ruid 1000) -> test binary (uid 0) -> fcvm -> firecracker

That privilege boundary breaks teardown, silently. When nextest's
slow-timeout.terminate-after fires it signals the test's process group, but the
kernel refuses a uid-1000 -> uid-0 signal, and killpg() still returns success
because it did signal one member -- the `sudo` front-end. SIGKILL cannot be
caught, so `sudo` never forwards it. The elevated test binary is orphaned to
init and keeps running, and because it never dies the pdeathsig chain BELOW it
(test binary -> fcvm -> firecracker) never fires either. Measured directly:
killpg(SIGKILL) from uid 1000 left 2/2 root processes alive, reparented to init.

On 2026-08-06 that left ~498 live firecracker processes on two ARM runners
(load average 523, 103 tasks in D-state, 420 zombies) for 21 hours, and the AWS
lease logic kept renewing the dead runners because GitHub still reported them
busy. Nothing in any job log ever mentioned them.

Teardown is per-hop: the kernel only kills a child whose OWN pdeathsig is set,
so one unprotected hop orphans the whole subtree below it. Every hop is now
covered.

The sudo hop
- scripts/root-test-runner.sh (new, 0755): `exec setpriv --pdeathsig KILL "$@"`.
  Kernel-enforced, so it holds in exactly the case that leaked -- a SIGKILL
  where no Drop impl, signal handler or `always()` cleanup step gets to run.
- Makefile: new ROOT_TEST_RUNNER variable replaces all 12 bare
  `sudo -E env PATH=$(PATH)` runner assignments (both triples x six invocation
  sites) across _test-root, _test-fc-mock, bench and bench-vm.

The test-binary hop
- tests/common/mod.rs: set_test_pdeathsig is now pub, and gains a
  std::process::Command flavour (set_test_pdeathsig_std); both share one
  pre_exec body (pdeathsig_hook) that sets PR_SET_PDEATHSIG(SIGKILL) and then
  closes the fork-vs-prctl race via getppid().
- Applied to all 16 hand-rolled fcvm spawns that bypassed spawn_fcvm*, across
  test_signal_cleanup, test_port_forward, test_proxy, test_localhost_image,
  test_serve, test_remap_file_range, test_fuse_in_vm_matrix and
  test_fuse_copy_file_range_vm. `fcvm serve` matters most: it owns every sandbox
  VM it creates, so orphaning it orphans all of them.

The VMM hop
- src/setup/rootfs.rs: the Layer-2 setup VM was the one VMM spawn in the tree
  that did not go through install_namespace_pre_exec, so a `fcvm setup` killed
  without unwinding (SIGKILL, a cancelled CI job) orphaned a live Firecracker
  that boots for minutes. It now uses the same helper (no namespaces -- the
  setup VM is on the host network -- so only the parent-death hook installs),
  plus kill_on_drop(true) for the several `?` paths that return before the wait
  loop.

Regression test
- tests/test_signal_cleanup.rs: test_root_test_runner_reaps_vm_when_sudo_is_killed
  reproduces the topology exactly -- the real runner script (resolved from
  CARGO_MANIFEST_DIR, so it cannot drift from the Makefile), a real sudo hop, a
  real booted VM -- then SIGKILLs ONLY the sudo front-end, the single process
  nextest can actually reach, and asserts both fcvm and firecracker are gone
  within 30s. It cleans up before asserting so a failure cannot itself leak a VM.

Defense in depth (NOT a substitute for the fix)
- .github/workflows/ci.yml: timeout-minutes on all three self-hosted jobs --
  Host 60, Host-Root 120, Container 60. GitHub's default is SIX HOURS, which is
  how a wedged runner held a job all day (one stuck >5h). Budgets are ~3-5x the
  longest of the last 6 successful runs, read from `gh api`: Host 12 min,
  Host-Root 43 min (arm64 SnapshotEnabled runs the suite twice), Container 21 min.
- scripts/ci-stray-vm-guard.sh (new, 0755) + pre/post steps in all three jobs:
  reports and reaps stray fcvm/firecracker/cloud-hypervisor processes at job
  boundaries, with a GitHub annotation when the count is non-zero so it lands on
  the run page instead of being buried in a step log. It matches on `comm`, not
  argv, so it can never match its own ps/grep. Limitation stated in its header:
  it only acts at job boundaries and cannot help a host that wedges mid-job
  (that remediation is AWS-side: terminate and replace), and D-state survivors
  need a reboot, not another kill.
  Zombies (stat Z) are excluded from BOTH the stray count and the post-SIGKILL
  survivor check, and reported separately. A zombie has already exited -- it
  holds no memory, no KVM fd, no vCPU thread -- so it is not a leaked microVM,
  and it can never be killed, so counting it would land it in SURVIVORS on every
  single run and fire the "survived SIGKILL, replace this runner" annotation for
  one stray <defunct>. A guard that cries wolf every run is a guard everyone
  learns to ignore. The 2026-08-06 incident had 420 zombies next to its 498 live
  firecrackers; only the live ones were the leak.
  A `--dry-run` flag reports without killing, so the script can be exercised
  somewhere other than a dedicated runner -- the default mode SIGKILLs every
  match, which on a shared box destroys someone else's VMs rather than garbage.
- .claude/CLAUDE.md: documents the per-hop rule, the three files that must keep
  covering their hop, and why a privilege boundary is special.

Verification (aarch64, real VMs)
- Both ways on the regression test, which is the whole point of the change:
    WITH the fix    PASS in 4.94s. Tree was sudo 743977 -> fcvm 743978 ->
                    firecracker 744257; SIGKILL to 743977 alone reaped both.
    WITHOUT the fix FAIL, twice (nextest retried): 35.18s and 34.84s, both with
                    "fcvm (PID 754949) survived the death of its sudo parent".
                    Reverted by reducing the runner to a bare `exec "$@"`.
  The test resolves the runner from CARGO_MANIFEST_DIR, so this exercises the
  same file the Makefile installs -- it cannot pass against a stale copy.
- make test-root FILTER=snapshot STREAM=1: 114 tests run, 114 passed [1222s],
  with the stray-process probe at 0 firecracker / 0 fcvm both before and after.
- make test-root FILTER=sanity: 5 tests run, 5 passed.
- make lint (fmt, clippy --all-targets -D warnings, cargo audit, cargo deny): ok
- make build: ok
- shellcheck on both new scripts: clean
- Stray-process probe around the sanity run: firecracker 0 -> 0, fcvm 0 -> 0.
  The failing negative run also left 0 behind, confirming the test cleans up
  before it asserts.
- ci-stray-vm-guard.sh exercised against live VMs in --dry-run: it correctly
  detected both `fcvm` and the 15-char-truncated `firecracker-def` comm, killed
  nothing, and exits 0 either way so it can never fail a job by itself.

Claude-Session: https://claude.ai/code/session_01QBekRuBULNUgUSGYmgkjyd

## Also found, NOT fixed here

These are wedge paths, not leak paths. With the pdeathsig chain intact a wedged
process is now killed when its parent dies, so they are no longer CI-fatal — but
they are real and worth separate PRs:

- `fcvm snapshot run` clones are **not children** of their `snapshot serve`: a
  clone self-registers `serve_pid` in its state file (`src/commands/snapshot.rs:811`),
  so serve's reaper (`src/commands/snapshot.rs:534-627`) is a best-effort state-file
  scan. A SIGKILL'd serve reaps nothing, and a clone still mid-restore is invisible
  to it because state is written only AFTER the firecracker spawn
  (`src/commands/common.rs:1502`).
- `src/commands/snapshot.rs:1604-1641`: the clone output-connect wait has no timeout
  and no cancel arm — it exits only if firecracker dies first.
- `src/commands/podman/mod.rs:1587` + `:470/:507` + `src/commands/snapshot.rs:1731`:
  on the snapshot cache-hit path a SIGTERM arriving before `cmd_snapshot_run`
  installs its own handler is consumed by an already-finished task and dropped,
  while tokio has already changed the process's SIGTERM disposition.
- `src/commands/common.rs:686-694`: cleanup aborts the VolumeServer
  (FUSE-over-vsock) tasks BEFORE `vm_manager.kill()`, whose `process.wait().await`
  is unbounded — the "guest blocked on a FUSE peer that is already gone" shape
  behind the incident's 103 D-state tasks.
- `src/uffd/server.rs:257`: the per-VM handler parks on `readable().await` with no
  timeout or cancellation, so `run()`'s post-cancel drain at `:191` is unbounded.
- `src/network/pasta.rs:604`: pasta is spawned with no pdeathsig.

---

## Commit 2 — `213fd4fa` Make the stray-VM guard's dry-run annotation tell the truth

The `::warning` annotation was emitted *before* the `--dry-run` branch, so a dry-run
announced "killing them" on the run page and then killed nothing. Moved the emit inside the
branch so each mode states what it actually did; reworded the zombie note so the phase reads
as a label. `shellcheck`: clean.

## Independent re-verification (second operator, quiet host, aarch64)

The run above was re-done from scratch on an idle host so the numbers are reproducible rather
than single-shot. Same conclusions, different PIDs:

| Step | Result |
|---|---|
| `make build` | exit 0 |
| `make lint` (fmt, clippy `--all-targets -D warnings`, audit, deny) | `advisories ok, bans ok, licenses ok, sources ok` |
| `make test-root FILTER=sanity` | `5 tests run: 5 passed, 679 skipped` |
| **POSITIVE** — fix in place | `1 test run: 1 passed` in 4.82s |
| **NEGATIVE** — runner reduced to `exec "$@"` | `1 test run: 0 passed, 1 failed`, both tries |
| `shellcheck` both scripts | clean |

The negative half fails for the **right reason** — the child is asserted *alive*, not merely
"the test failed":

```
thread 'test_root_test_runner_reaps_vm_when_sudo_is_killed' panicked at tests/test_signal_cleanup.rs:829:5:
fcvm (PID 857921) survived the death of its sudo parent — the privileged test runner is not
setting PR_SET_PDEATHSIG, so a nextest timeout kill orphans the whole VM tree
```

Positive run, for contrast — tree `sudo 848847 -> fcvm 848848 -> firecracker 849058`, SIGKILL
to `848847` alone, both descendants gone inside the 30s bound.

The same asymmetry reproduces on the bare primitive in ~10s, with no VM involved, which is
useful when triaging this in future:

```
NEGATIVE (bare 'exec "$@"'):        CHILD SURVIVED — state=S, reparented to ppid=1
POSITIVE (setpriv --pdeathsig KILL): CHILD REAPED BY KERNEL
```

**Stray-process probe:** `live firecracker=0 live fcvm=0` before; `0 / 0` after, on a
re-check. The immediate post-probe of the *negative* run caught `firecracker=1` mid-teardown
— the test issues its cleanup kill before asserting, so that is the kill in flight, and it
drains to 0. Worth stating rather than rounding down to a clean "0".

**Guard behaviour, both modes, against synthetic `fcvm`/`firecracker` processes** (the guard
matches on `comm`, so `cp /bin/sleep .../fcvm` is a faithful stray):

```
$ ./scripts/ci-stray-vm-guard.sh pre --dry-run
::warning title=Stray microVMs (pre, dry-run)::1 ... NOT killing them (--dry-run).
  synthetic pid 865440 ALIVE (correct)

$ ./scripts/ci-stray-vm-guard.sh post
::warning title=Stray microVMs (post)::1 ... killing them.
killed 1, still present after SIGKILL: 0
  synthetic pid GONE (correct)
```

Zombie exclusion, checked against a synthetic `ps` table so every state is covered — `Z` and
`Z+` dropped, `D` (the actual wedge signal) still caught, non-VM processes never matched:

```
kept:    111 Sl firecracker-def | 333 Sl fcvm | 444 D firecracker
dropped: 222 Z fcvm | 555 Z+ firecracker-def | 666 sshd (never matched)
```

### Two process-safety notes from doing this verification

* **The guard is destructive by default and has no interlock against a concurrent job.** I ran
  it on a shared dev box to exercise it and it SIGKILLed a live, unrelated test's VM one second
  after that test started. On a dedicated CI runner (one job at a time) that is correct and
  intended; anywhere else `--dry-run` is mandatory. This is exactly why the flag was added.
* **`kill -0` is not a liveness check across a privilege boundary.** A first cut of the
  primitive probe used `kill -0 <root pid>` from uid 1000, which returns `EPERM`, not `ESRCH` —
  so a *surviving* root child read as "reaped" and the negative case looked like it passed.
  Liveness here must come from `/proc/<pid>/stat` (and must exclude state `Z`). That is the same
  uid-1000→uid-0 signal asymmetry this PR is about, which makes it an easy way to fool yourself
  while testing the fix.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -981,6 +981,38 @@ assert!(localhost_works, "Localhost port forwarding should work (requires route_
 
 Do not run `sudo cargo`. Use Makefile targets so cargo runs as your user and test binaries run via `CARGO_TARGET_*_RUNNER='sudo -E'`; otherwise `target/` becomes root-owned and breaks builds.
 
+### PROCESS TEARDOWN IS PER-HOP. EVERY HOP.
+
+**`PR_SET_PDEATHSIG` only kills a child whose OWN pdeathsig is set. One unprotected hop
+orphans the entire subtree below it.** The chain that must hold, end to end:
+
+```
+cargo-nextest (uid 1000) → sudo (ruid 1000) → test binary (uid 0) → fcvm → firecracker
+                           ^ setpriv --pdeathsig  ^ set_test_pdeathsig  ^ install_namespace_pre_exec
+```
+
+- `scripts/root-test-runner.sh` (`ROOT_TEST_RUNNER` in the Makefile) covers the **sudo hop**.
+  Never reduce the privileged runner back to a bare `sudo -E env PATH=...`.
+- `tests/common/mod.rs::set_test_pdeathsig{,_std}` covers **every long-lived fcvm spawn** in
+  tests. `spawn_fcvm*` applies it; a hand-rolled `Command::new(&fcvm_path)…spawn()` must too.
+- `src/utils.rs::install_namespace_pre_exec` covers **every VMM spawn** (Firecracker, Cloud
+  Hypervisor, and the Layer-2 setup VM in `src/setup/rootfs.rs`). It must stay the LAST
+  `pre_exec` — `setns(CLONE_NEWUSER)` zeroes `pdeath_signal`, so setting it earlier is lost.
+
+**Why a privilege boundary is special:** the kernel refuses a signal from uid 1000 to a uid-0
+process, and `killpg` still returns success when it managed to signal *any* member — so
+nextest's slow-timeout kill fails **silently** against a `sudo`'d test binary. SIGKILL cannot
+be caught, so `sudo` never forwards it either. On 2026-08-06 that left ~498 `firecracker`
+processes alive on two ARM runners (load average 523, 103 tasks in D-state, 420 zombies) for
+21 hours, and the AWS lease logic kept renewing the dead runners because GitHub still reported
+them busy. Regression test: `test_root_test_runner_reaps_vm_when_sudo_is_killed`
+(`tests/test_signal_cleanup.rs`). Defense in depth (NOT a substitute): `timeout-minutes` on
+every self-hosted job plus the `scripts/ci-stray-vm-guard.sh` pre/post steps in `ci.yml`.
+
+**Prefer kernel-enforced reaping over cleanup code.** A `Drop` impl, a signal handler, or an
+`always()` cleanup step does not run when the process is SIGKILLed — which is exactly the case
+that leaked. If teardown matters, it must survive SIGKILL.
+
 ### Container Build Rules
 
 **Container builds work naturally with layer caching.** No workarounds needed.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,6 +228,11 @@ jobs:
     needs: [skip-check]
     if: ${{ github.event.inputs.job != 'container' && needs.skip-check.outputs.skip != 'true' }}
     runs-on: [self-hosted, Linux, '${{ matrix.arch }}']
+    # Longest of the last 6 successful runs: 12 min. GitHub's default is SIX HOURS, which is
+    # how a wedged runner held a job all day on 2026-08-06 (one job stuck >5h) while the
+    # AWS-side lease logic kept renewing its lease because GitHub still reported it busy.
+    # 5x headroom covers a cold runner (kernel/rootfs rebuilds) and is nowhere near 6h.
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
@@ -259,6 +264,13 @@ jobs:
       - name: Disk preflight
         working-directory: fcvm
         run: ./scripts/runner-disk-preflight.sh
+      # A self-hosted runner is reused forever, so a leaked microVM is inherited by every
+      # later job. Report (and reap) what the PREVIOUS job left behind before doing anything
+      # else — on 2026-08-06 two runners had accumulated ~498 live firecracker processes and
+      # nothing in any job log had ever mentioned them.
+      - name: Stray microVM guard (pre)
+        working-directory: fcvm
+        run: ./scripts/ci-stray-vm-guard.sh pre
       - uses: ./fcvm/.github/actions/checkout-deps
       - name: Install Rust
         run: |
@@ -412,6 +424,12 @@ jobs:
       - name: test-fast
         working-directory: fcvm
         run: make test-fast
+      # And what THIS job left behind. A non-zero count here means the run leaked microVMs:
+      # it is annotated on the run page, not just buried in the step log.
+      - name: Stray microVM guard (post)
+        if: always()
+        working-directory: fcvm
+        run: ./scripts/ci-stray-vm-guard.sh post
       - name: Upload test logs
         if: always()
         uses: actions/upload-artifact@v6
@@ -446,6 +464,10 @@ jobs:
     needs: [skip-check]
     if: ${{ github.event.inputs.job != 'container' && needs.skip-check.outputs.skip != 'true' }}
     runs-on: [self-hosted, Linux, '${{ matrix.arch }}']
+    # Longest of the last 6 successful runs: 43 min (arm64 SnapshotEnabled, which runs the
+    # whole suite twice, plus the nested + btrfs kernel builds). ~3x headroom.
+    # See the Host job comment for why the 6h default is not acceptable here.
+    timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:
@@ -487,6 +509,13 @@ jobs:
       - name: Disk preflight
         working-directory: fcvm
         run: ./scripts/runner-disk-preflight.sh
+      # A self-hosted runner is reused forever, so a leaked microVM is inherited by every
+      # later job. Report (and reap) what the PREVIOUS job left behind before doing anything
+      # else — on 2026-08-06 two runners had accumulated ~498 live firecracker processes and
+      # nothing in any job log had ever mentioned them.
+      - name: Stray microVM guard (pre)
+        working-directory: fcvm
+        run: ./scripts/ci-stray-vm-guard.sh pre
       - uses: ./fcvm/.github/actions/checkout-deps
       - name: Install Rust
         run: |
@@ -623,6 +652,12 @@ jobs:
         run: |
           make clean-test-data || true
           make bench-vm
+      # And what THIS job left behind. A non-zero count here means the run leaked microVMs:
+      # it is annotated on the run page, not just buried in the step log.
+      - name: Stray microVM guard (post)
+        if: always()
+        working-directory: fcvm
+        run: ./scripts/ci-stray-vm-guard.sh post
       - name: Capture kernel logs
         if: always()
         run: |
@@ -650,6 +685,9 @@ jobs:
     needs: [skip-check]
     if: ${{ github.event.inputs.job != 'host' && needs.skip-check.outputs.skip != 'true' }}
     runs-on: [self-hosted, Linux, '${{ matrix.arch }}']
+    # Longest of the last 6 successful runs: 21 min. ~3x headroom.
+    # See the Host job comment for why the 6h default is not acceptable here.
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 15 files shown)


ejc3:
## Independent verification (second agent, same commit)

Two agents verified this branch separately, without sharing intermediate results — the coordinator had launched the task twice. Same both-ways outcome, so the evidence below independently reproduces the numbers already on this PR rather than restating them.

All runs on aarch64 against real booted microVMs, at `scripts/root-test-runner.sh` blob `12c982ee` — byte-identical to the copy on this PR head (`b2b566c0`).

### The both-ways test

| | Result |
|---|---|
| **With the fix** | **PASS**, 4.94s. Tree was `sudo 743977 -> fcvm 743978 -> firecracker 744257`; SIGKILL to `743977` alone reaped both descendants. |
| **Fix reverted** (runner cut down to a bare `exec "$@"`) | **FAIL**, twice — nextest retried: 35.18s and 34.84s. Both: `fcvm (PID 754949) survived the death of its sudo parent — the privileged test runner is not setting PR_SET_PDEATHSIG, so a nextest timeout kill orphans the whole VM tree` |

Two details worth recording, because they show the test is measuring the real thing and not a proxy:

* In the passing run the process directly under `sudo` is **`fcvm` itself** (`parent chain: [743978, 743977]`). `env -> sh -> setpriv -> fcvm` all `exec` in place, so `PR_SET_PDEATHSIG` is armed against the `sudo` front-end with no intermediate process to absorb it. That is exactly the CI topology.
* In the failing run an **unrelated** `firecracker` (PID 752607, parent chain terminating at 7354) was live on the box from another workload. The test's ancestry check correctly rejected it (`is_descendant_of(754949) = false`) and selected only its own child 755242 — so the assertion cannot be satisfied or defeated by ambient processes.

### Supporting runs

* `make test-root FILTER=sanity` — 5 tests run, **5 passed**.
* `make lint` (fmt, `clippy --all-targets -D warnings`, `cargo audit`, `cargo deny`) — clean.
* `make build` — clean.
* `shellcheck` on both new scripts — clean.
* Stray-process probe around the sanity run: `firecracker` 0 → 0, `fcvm` 0 → 0. The **failing** negative run also left 0 behind, which confirms the test's "clean up before asserting" ordering actually holds — a failure of this test cannot itself leak a microVM.

Worth noting separately: an unrelated `make test-root FILTER=snapshot` run (~20 min, full snapshot suite) completed on this same box through the new runner and left **0** `fcvm`/`firecracker` processes behind, so `setpriv` is not interfering with normal teardown on the success path.

### On `ci-stray-vm-guard.sh`

I exercised the guard against live VMs with the kill path stubbed out. It correctly detected both `fcvm` and the 15-char-truncated `firecracker-def` comm, and exits 0 in every path so it can never fail a job by itself.

I also independently hit the zombie problem that `daf39098` now fixes: a baseline probe on this box showed **12 zombie `fcvm` processes**. Under the pre-fix matcher those would have been counted as strays, been unkillable (SIGKILL to a zombie is a no-op), and fired the "survived SIGKILL, this runner needs a reboot" annotation on every single run. The 2026-08-06 incident had 420 zombies alongside the real leak, so excluding them is the difference between a usable signal and a permanent false alarm. Good catch.

### Scope check

`_test-all` — the target the Container job runs inside the container — goes through `scripts/no-sudo.sh`, not the privileged runner, so nothing in this change adds a `setpriv` dependency to the container image. The runner is only used on the bare-metal Host/Bench runners, where `setpriv` ships in `util-linux` (Priority: required).

One property that makes the pre-job guard safe: runners register one runner process per instance (`--name runner-$INSTANCE_ID` in `setup-runner.sh`), so a job's `pre` guard cannot reap a *concurrent* job's live VMs on the same box.

### Agreeing with the stated limitation

The guard acts only at job **boundaries**. It reports what the previous job left and what this job left; it does not prevent a host from wedging **mid-job**, and D-state survivors need a reboot rather than another kill. Remediation for an already-wedged host is an AWS-side terminate-and-replace, not this script. The guard is instrumentation and cleanup — `setpriv --pdeathsig KILL` is the actual fix.

PR #731, opened 2026-08-08
bench: withdraw the CDP A/B numbers and fix the harness defects that invalidated them

ejc3:
> **2026-08-08 update after adversarial review: every latency and CPU figure this
> PR originally published is WITHDRAWN.** They were not wrong measurements of the
> right thing — they were measurements taken through a harness with defects that
> make them undefendable. The withdrawals and their specific reasons are recorded
> in [`bench/chromium/REVIEW.md`](bench/chromium/REVIEW.md); the harness defects
> are fixed here with tests that were watched fail first. **This PR now ships the
> corrected harness and the withdrawal ledger. It publishes no A/B result.**

## What this PR is now

Harness, container, and documentation only — still no `src/` change. Three things:

1. **A regression test that could not fail is repaired** (highest priority).
2. **Every published CDP figure is withdrawn**, with the defect that invalidated it.
3. **16 harness defects fixed**, each with a red-verified test.

## 1. The assertion that could not fail

`test_bench_fast_teardown_leaks_nothing_clone` held the namespace holder as a bare
`Option` and asserted:

```rust
let holder = find_holder_for_fcvm(clone_pid);          // Option<Tracked>
let holder_alive = holder.as_ref().is_some_and(|h| h.running());
assert!(!holder_alive, "LEAK: namespace holder survived ...");   // !false when None
```

When discovery returns `None` this is `assert!(!false)`. Discovery is coupled to
production by a literal string — `find_holder_for_fcvm` pgreps for `sleep
infinity`, which `PastaNetwork::build_holder_command` emits — so any argv drift
disarmed it silently. It never looked for pasta at all, and its cleanup was a
no-op on the same path, so the test guarding against leaks would itself leave an
orphaned holder pinning a user+net namespace.

**Red-verified by fault injection** (`build_holder_command` argv →
`sleep 2147483647`, everything else unchanged):

| step | code | test | result |
|---|---|---|---|
| 1 | unmutated | fixed | **passed** |
| 2 | mutated | **old** | **passed** — `firecracker PID 2959733, holder None` |
| 3 | mutated | fixed | **failed** — `Error: clone has no namespace holder child — … If discovery drifts, the holder-leak assertion below checks nothing at all` |
| 4 | reverted | fixed | **passed** |

Step 2 is the finding: the old test passed while reporting `holder None`.

Now all three children are hard-fail-on-absent, the fixture is torn down before
any early return, survivors are listed and killed before asserting, and the data
dir is asserted to survive the SIGKILL alongside the state file. A permanent drift
alarm (`holder_discovery_pattern_tracks_production_holder_argv`) pins the pgrep
pattern to the production argv.

## 2. Withdrawn figures

| withdrawn | why it cannot stand |
|---|---|
| `cdp` **384 ms**, `cdp-fast` **372 ms**, `PART 1 −180.5 ms` | Success-conditioned with an arm-correlated censoring rate the analyzer physically could not report — one global `n_failed`, no per-arm denominator anywhere. CDP dropped **1/60 file (1.7%)** and **6/60 UFFD (10.0%)** with `WsClosed`; `exec` dropped **0/66**. A ~5.25 s transport drop truncates the right tail, so the 10% arm is pulled down ~6× harder than the 1.7% arm — in the same direction as the reported effect. |
| reclaim CPU **0.00 [0.00, 0.00] ms** | Three defects in one number: `/proc/<pid>/stat` is jiffy-quantized (10 ms here) so a sub-tick reclaim reports an exact zero with zero uncertainty; the analyzer **pooled all children and discarded the name**, so the median of {firecracker, holder, pasta} medians the straggler away; and the sampler ran **sequentially**, recording `null` for any child that exited during an earlier child's sample. Honest restatement: *below /proc tick resolution (< 20 ms) for every child sampled*. |
| machine cost **+610.4 ms [+575.9, +620.3]** | The "ambient" baseline was measured while the harness held 100% of one core (`while time.monotonic()-t0 < 0.05: pass`), then multiplied by the whole reclaim window and subtracted. Measured on this box, same shape back to back: **spinning 1.40 / 1.40 / 1.20 cores** vs **sleeping 0.20 / 0.40 / 0.00**. The control window was also taken *before* the kill, so it additionally contains the live VM's own CPU. |
| per-child **pasta 704 ms**, and "the pdeathsig guarantee does not cover pasta" | Measured on a tree predating `46dbb789`, which arms `PR_SET_PDEATHSIG(SIGKILL)` on pasta. Both withdrawn. Replacement claim: kernel-enforced for the VMM and holder **unconditionally**, and for pasta **while fcvm runs as root** (`cred_cap_issubset` holds only under a capability loss). |
| "Early response converts teardown from latency into throughput cost" — **REFUTED** | The refutation rested on the two rows above. Both withdrawn ⇒ the premise returns to **untested**, not to supported. |

**Availability gate for the re-run:** ≥200 CDP requests per backend at 0 failures
before any CDP latency figure is quoted (0/200 ⇒ CP upper bound **1.8%** — `reqanalyze.clopper_pearson(0, 200)` = [0.000%, 1.828%]; the 1.5% previously quoted here and in REVIEW.md was the ONE-sided bound, inconsistent with this file's declared two-sided convention). Today's
observed rates fail that loudly.

`AGENTS.md` also gains uncertainty and `n` on every reference row, quotes the
restore primitive **once** (~6 ms — it said 5.8 ms in a table and 4 ms in the
prose), marks the teardown table as 10 ms-quantized with its slope requoted as a
fitted range, replaces the "cannot leak" heading with the scoped guarantee, and
reconciles its REFUTED list with `REVIEW.md` instead of contradicting it.

## 3. Harness defects fixed

**`reqbench.py`** — `teardown_fast` refuses to reap the state file and data dir of
a VM it failed to kill (it used to record `all_gone: false` and then `rmtree` the
rootfs out from under a live Firecracker), SIGKILLs survivors, and aborts the
schedule; control window sleeps and harness self-CPU is subtracted from both
windows; per-child CPU sampled concurrently and emitted as an interval with
`below_resolution`; `find_state` is inotify-driven with the watch registered
*before* the spawn and matches on name as well as pid (it was a no-sleep
`listdir`+`json.load` spin burning a full core for its whole duration, next to a
restoring 2-vCPU clone); the exec arm's bare `proc.wait(timeout=…)` no longer
raises out of `main()` and orphans a clone; `--ws-url` is re-hosted onto each
clone's own endpoint; same-comm children can no longer collide and drop one.

**`reqanalyze.py`** — per-arm `attempted`/`failed` with exact Clopper-Pearson
intervals and a `publishable` gate; **the teardown/leak section reads every
attempt, not the ok-filtered list**, so `all_gone: false` on a *failed* request is
no longer invisible; per-child CPU reported by comm; cross-arm deltas labelled
CENSORED.

**`reqbench.sh`** — trap-based cleanup with `$!` captured in every phase (two
captured nothing at all); `cmd_verify` returns a failure count instead of `|| echo`
swallowing all three hops and exiting 0; target-id stability compared across **two**
clones instead of printing one id and asserting nothing.

**`cdpdrive.py` / `render.py` / `teardown_probe.py`** — `failure_class` for
transport drops; `--print-target` implemented (its docstring had promised it since
the file was written); `_recv_until` keeps bytes past the handshake marker, whose
loss desyncs the next frame header and surfaces as exactly the observed
`WsClosed("connection closed mid-frame")` — **a variable removed, not a
diagnosis**; `teardown_probe`'s `no state` path no longer leaks the entire clone
and now reaps the data dir.

## Test evidence

```
$ make lint
fmt, clippy -D warnings, audit, deny: clean

$ make test-root FILTER="-E 'test(/holder_discovery_pattern|test_bench_fast_teardown_leaks_nothing_clone/)'"
clone tree: firecracker=2990188, holder=2990182, pasta=2990191
test_bench_fast_teardown_leaks_nothing_clone PASSED
Summary [1.811s] 2 tests run: 2 passed, 689 skipped

$ python3 -m unittest test_reqbench       # bench/chromium/test_reqbench.py
Ran 13 tests in 8.6s — OK
```

All 13 Python tests were watched fail against the unmodified pre-fix files (12
failed/errored; the 13th — the sampling-order test — was made deterministic and
red-verified separately against the untouched sequential sampler).

**Not red-verified, and why:** the `WsClosed` root cause (needs a live clone, fires
at 1.7–10%, no deterministic reproduction); `cmd_verify`'s non-zero exit and the
two-clone target-id comparison (shell paths needing a golden snapshot + serve —
mechanical and `bash -n` clean, but no full verify cycle was run); the pasta arm of
the clone test's discovery (pasta was found on every run, so that `.context()` has
not been observed firing).

Rebased onto `origin/main` to pick up `46dbb789` (pasta pdeathsig) and `3ee2e4f6`
(its root-only precondition).

## Summary by CodeRabbit

- **New Features**
  - Added Chromium benchmarking workflows for CDP, execution, teardown, and control-path comparisons.
  - Added screenshot driving with navigation timing, validation, retries, and JSON output.
  - Added health checks for Chromium readiness and CDP connectivity.
  - Added analysis tools for benchmark results, execution stages, teardown, and confidence intervals.
  - Added automated build, verification, contention, and cleanup workflows.

- **Bug Fixes**
  - Improved WebSocket handling to preserve data received during connection setup.

- **Documentation**
  - Added benchmark methodology, networking, health-check, and performance guidance.

- **Tests**
  - Added integration and regression coverage for teardown, cleanup, accounting, and request failures.

### DIFF
--- a/Containerfile.chromium-bench
+++ b/Containerfile.chromium-bench
@@ -6,7 +6,17 @@
 # each drive one CDP render via bench/chromium/render.py.
 #
 # Build (repo root context; .dockerignore excludes target/):
-#   podman build -t localhost/chromium-bench -f Containerfile.chromium-bench .
+#   podman build --format docker -t localhost/chromium-bench -f Containerfile.chromium-bench .
+#
+# --format docker is LOAD-BEARING. podman's default OCI format DROPS the
+# HEALTHCHECK at the bottom of this file with only a warning
+# ("HEALTHCHECK is not supported for OCI image format and will be ignored")
+# and exits 0. fcvm then sees no healthcheck, and src/health.rs treats a
+# MISSING one as a PASS (`None => HealthStatus::Healthy`), so the golden
+# snapshot fires on "container is running" — a COLD browser — and every
+# restored clone's "warm" latency is really a first-paint number. Verify:
+#   podman inspect localhost/chromium-bench --format '{{json .HealthCheck}}'
+# must print the Test array containing cdp_health.
 #
 # Host smoke test (fast loop, no VM):
 #   podman run -d --name cb localhost/chromium-bench
@@ -19,9 +29,12 @@
 #   fcvm exec --pid <PID> -c -- python3 /opt/bench/render.py \
 #       http://127.0.0.1:8000/medium.html --out-prefix /tmp/medium
 #
-# Golden-snapshot warm point: /ready returns 200 only after warmup, so
-#   --health-check http://127.0.0.1:8000/ready
-# makes the auto startup snapshot capture a WARM browser.
+# Golden-snapshot warm point: SUPERSEDED — do NOT pass a --health-check URL.
+# An earlier recipe here suggested `--health-check http://127.0.0.1:8000/ready`.
+# That sets `has_http_check` in src/health.rs, which SKIPS the podman HEALTHCHECK
+# entirely — i.e. it disables the CDP round-trip gate defined at the bottom of
+# this file, which is the trigger this design actually relies on. Leave
+# --health-check UNSET so fcvm's AND-logic consults the image's HEALTHCHECK.
 #
 # Fixtures live in bench/chromium/pages/ and are COPY'd in; the host-side
 # fixture server (bench harness) serves the SAME directory so in-guest and
--- a/Containerfile.chromium-bench
+++ b/Containerfile.chromium-bench
@@ -40,6 +40,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     chromium \
     python3 \
     fonts-dejavu-core \
+    socat \
     && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/partial
 
 # Belt-and-braces half of the getenv/setenv workaround documented at length in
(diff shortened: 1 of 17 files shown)


ejc3:
Rebased onto current main and force-pushed (`39602de9` → `6bd95bbb`). Now MERGEABLE.

**The one conflict was `bench/chromium/REVIEW.md`, resolved in favour of main.** Not a judgement call about which document is better written — they describe different runs, and one is explicitly retired by the other:

- main's version is the ledger for the **corrected** run (`results/20260808-corrected`), and it opens with "Two runs are on record. **Quote only from the second.**"
- this branch's version was the review of the **first** run (`20260807-full`), whose comparative numbers were retracted on six methodology defects.

main's ledger already carries the first run in its table (`comparative numbers retracted on methodology`) and tracks each previously-refuted claim through to its post-correction status. The rerun requirements at the bottom of this branch's version were "binding for the next full run" — that run has since happened, so they are discharged, not lost. Keeping the superseded document alongside the one that supersedes it is how a stale number gets quoted six months from now.

Verified the resolution rather than assuming it: the resolved file is byte-identical to `origin/main:bench/chromium/REVIEW.md`, which is why REVIEW.md no longer appears in this PR's diff at all (13 files → 12).

**This PR's actual contribution is untouched** — `reqbench.py` (646), `cdpdrive.py` (320), `reqanalyze.py` (275), `reqbench.sh` (221), `reqstages.py` (130), `teardown_probe.py` (124), `cdp_health.py` (75), the `entry.sh` changes, and `tests/test_signal_cleanup.rs` (176).

Checks run locally before pushing, since a rebase can produce a tree nobody has ever compiled:
- every `bench/chromium/*.py` passes `python3 -m py_compile`; every `*.sh` passes `bash -n`
- the two `fcvm::` paths the new Rust test depends on both exist on main — `utils::graceful_kill` (`src/utils.rs:131`) and `paths::state_dir` (`src/paths.rs:122`). Worth checking explicitly: #736 was caught today calling a helper that existed only on a branch I had closed, so "it compiled on my base" is not evidence it compiles on main.
- `git merge-tree` against #735 shows **no** conflict, even though both append to `tests/test_signal_cleanup.rs` — they land in different regions, so merge order between the two does not matter.

#741 is stacked on this branch and should have its base retargeted to main once this merges.

PR #735, opened 2026-08-08
Overlap VM teardown kills instead of serializing them; arm pdeathsig on pasta

ejc3:
Teardown ran its kills back to back even though nothing orders them. This overlaps
them, and arms `PR_SET_PDEATHSIG` on pasta — the one hop in the rootless tree that
did not have it.

## Fast teardown

~154ms of every clone request happens after the result exists. `cleanup_vm()` awaited
each kill before sending the next, so the VMM's address-space reclaim (~40ms of
`exit_mmap()` on a ~1GiB guest mapping), the namespace holder's exit and pasta's
teardown (~28ms) serialized. Restructured into four phases:

1. **Signal, block on nothing** — SIGKILL the VMM, the holder and the network helper,
   and cancel/abort the in-process tasks, with no `await` in between.
2. **Join the waits together** — teardown costs the slowest exit, not the sum.
3. **Network cleanup** — once nothing is left using the namespace.
4. **On-disk reaping** — state file, NFS exports, Firecracker log, data directory.

Phase 4 is **fully synchronous**: `cleanup_vm` does not return until all of it is
gone. No janitor, no deferral. "Leave nothing on disk" is a correctness contract just
like "leave no orphaned VM" — a deferred sweep lets a crash strand a state file whose
PID is then reused, a reflinked rootfs, or an `/etc/exports.d` entry that makes every
later `exportfs -ra` fail. The three steps are independent so they run concurrently
(the log copy and directory removal share a future, since the log lives in the
directory), but all inside the blocking window.

Enabling that required splitting the kill in two:

- `Hypervisor` gains `fn start_kill()` and `async fn reap()`. `kill()` becomes a
  provided method composing them, so both backends keep a single implementation.
- `NetworkManager` gains `fn start_kill_processes()`, default no-op, implemented by
  `PastaNetwork`. `cleanup()` still kills *and* reaps unconditionally, so it stays
  correct on the error paths that never call `start_kill_processes`.

### Reporting it honestly

Teardown emits one record with three **distinct** numbers, so overlapping work can
never be presented as eliminating it:

| field | meaning |
|---|---|
| `caller_blocking_ms` | what the request actually pays, start to finish |
| `until_gone_ms` | first SIGKILL until the last resource is released |
| `reclaim_cpu_ms` | CPU the kernel burned destroying the processes |

plus `processes_us` / `network_us` / `disk_us` (microseconds — two phases are
sub-millisecond and would round to a useless 0).

`reclaim_cpu_ms` needed a correction *during* this work. `RUSAGE_CHILDREN` credits a
child's **entire lifetime** CPU at the moment it is reaped, so the first version
reported `reclaim_cpu_ms=12328` for a VM that had simply been alive a while. It is now
the delta minus the CPU the VMM and holder had already spent (sampled from
`/proc/<pid>/stat` just before the signal), sampled after phase 2 so pasta's lifetime
CPU cannot leak in. What remains is genuinely the post-SIGKILL cost.

Observed on real clone teardowns:

```
caller_blocking_ms=41  until_gone_ms=41  processes_us=35832 network_us=143 disk_us=5716
caller_blocking_ms=135 until_gone_ms=135 processes_us=77085 network_us=85  disk_us=58390
```

`caller_blocking` and `until_gone` being equal is the point: nothing was hidden.

## pasta pdeathsig — and a correction to the severity

pasta is now spawned with `PR_SET_PDEATHSIG(SIGKILL)` via `pre_exec`, plus a
`getppid()` re-check that fails the exec if the parent died inside the fork/exec
window. AGENTS.md requires teardown be armed per hop; pasta was the only hop without
it.

**It was graded CRITICAL on the theory that an orphaned pasta keeps a recycled
loopback port bound and misroutes a client into an old clone's namespace. That does
not reproduce, and this PR does not claim it does.** Measured on this branch:

- With the `pre_exec` **removed** and the binary verifiably rebuilt, pasta still died
  on fcvm SIGKILL. The reaping test passed either way.
- Killing only the holder (fcvm alive) left pasta in state `Z` — it had exited. So the
  trigger is the holder's death, not fcvm's.
- Repeating the SIGKILL with the holder's netns **pinned open by an external fd** still
  killed pasta, ruling out namespace teardown. pasta joins the holder *by PID* and
  passt exits when that PID exits.

So pasta cannot currently outlive its holder, and the holder cannot outlive fcvm. The
fix is still worth its ten lines: today's guarantee runs through two things fcvm does
not own — the holder's own pdeathsig, and an **undocumented** passt behaviour
(`--no-netns-quit` documents only the filesystem-bound case) in a binary this repo
pins to a patched fork and periodically rebases. This makes pasta's death depend on
nothing but the kernel.

## Test

`test_sigkill_kills_firecracker_rootless` → `test_sigkill_reaps_rootless_vm_tree`
(extended, not duplicated):

- finds firecracker, the holder **and** pasta, and **requires all three to be found** —
  previously a missing holder or pasta silently skipped its own assertion, exactly the
  hedge that lets an orphan go unnoticed;
- asserts none survive, reporting every survivor in one message and killing them first
  so a failure of this test cannot itself leak a microVM;
- documents its scope honestly: it guards the **end state**, not pasta's pdeathsig
  specifically, which no VM-level test can isolate for the reason above.

## Test Results

```
$ make lint
fmt, clippy --all-targets -D warnings, audit, deny — all clean (exit 0)

$ make build
Finished `release` profile

$ make test-root FILTER=sigkill STREAM=1
Our tree: firecracker=1744507, holder=1744435, pasta=1744511
Sending SIGKILL to fcvm (PID 1744433)
firecracker, holder and pasta all reaped by the kernel
test_sigkill_reaps_rootless_vm_tree PASSED
     Summary [5.343s] 1 test run: 1 passed, 687 skipped
```

`make test-root FILTER=snapshot` is queued to run on a free box — the host had a
concurrent VM suite from another worktree writing the same shared `FCVM_DATA_DIR`,
and results from a contended box are not worth reporting. Will update this PR with
the outcome.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved virtual machine shutdown reliability by signaling processes promptly and completing cleanup more efficiently.
  * Prevented orphaned networking and namespace processes when a virtual machine exits unexpectedly.
  * Ensured Firecracker, networking helpers, logs, disk data, NFS exports, and related resources are fully cleaned up.
  * Added safeguards to verify that all shutdown-related processes terminate correctly.
  * Improved shutdown monitoring with timing and CPU-reclaim measurements.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -987,8 +987,10 @@ Do not run `sudo cargo`. Use Makefile targets so cargo runs as your user and tes
 orphans the entire subtree below it.** The chain that must hold, end to end:
 
 ```
-cargo-nextest (uid 1000) → sudo (ruid 1000) → test binary (uid 0) → fcvm → firecracker
-                           ^ setpriv --pdeathsig  ^ set_test_pdeathsig  ^ install_namespace_pre_exec
+cargo-nextest (uid 1000) → sudo (ruid 1000) → test binary (uid 0) → fcvm ─┬→ firecracker
+                           ^ setpriv --pdeathsig  ^ set_test_pdeathsig     ├→ holder
+                                                                           └→ pasta (rootless)
+                                                                              ^ see per-hop list below
 ```
 
 - `scripts/root-test-runner.sh` (`ROOT_TEST_RUNNER` in the Makefile) covers the **sudo hop**.
@@ -998,6 +1000,14 @@ cargo-nextest (uid 1000) → sudo (ruid 1000) → test binary (uid 0) → fcvm �
 - `src/utils.rs::install_namespace_pre_exec` covers **every VMM spawn** (Firecracker, Cloud
   Hypervisor, and the Layer-2 setup VM in `src/setup/rootfs.rs`). It must stay the LAST
   `pre_exec` — `setns(CLONE_NEWUSER)` zeroes `pdeath_signal`, so setting it earlier is lost.
+- `src/commands/common.rs::spawn_namespace_holder` covers the **holder hop** (rootless mode).
+  The holder's pdeathsig is armed in its `pre_exec`, before UID/GID mappings are written.
+- `src/network/pasta.rs` (in `start_pasta`) covers the **pasta hop** (rootless mode). Must be
+  the last `pre_exec`; includes a `getppid` re-check to close the fork/exec race window.
+
+Regression tests: `test_sigkill_reaps_rootless_vm_tree` asserts all three children (firecracker,
+holder, pasta) die when fcvm is SIGKILL'd. `test_root_test_runner_reaps_vm_when_sudo_is_killed`
+covers the sudo hop specifically.
 
 **Why a privilege boundary is special:** the kernel refuses a signal from uid 1000 to a uid-0
 process, and `killpg` still returns success when it managed to signal *any* member — so
--- a/src/commands/common.rs
+++ b/src/commands/common.rs
@@ -637,16 +637,95 @@ pub struct CleanupContext {
     pub output_listener_handle: Option<JoinHandle<Vec<(String, String)>>>,
 }
 
-/// Cleanup resources for a VM (used by both podman and snapshot commands)
+/// How long teardown waits for the health monitor to observe its cancellation before
+/// moving on. It only writes state through the per-VM flock, which `delete_state` takes
+/// too, so an unjoined straggler is safe — this is a courtesy join, not a barrier.
+const HEALTH_MONITOR_STOP_BUDGET: std::time::Duration = std::time::Duration::from_millis(100);
+
+/// Total CPU charged to this process's REAPED children so far (`RUSAGE_CHILDREN`).
 ///
-/// This function handles the complete cleanup sequence:
-/// 1. Cancel health monitor gracefully
-/// 2. Abort volume server tasks
-/// 3. Kill VM process
-/// 4. Kill holder process (rootless mode)
-/// 5. Cleanup network resources
-/// 6. Delete state file
-/// 7. Remove data directory
+/// A child's ENTIRE lifetime CPU lands here at the moment it is reaped, not as it runs — so
+/// a raw delta across teardown is the VMM's whole-life CPU (measured: 12.3 CPU-seconds for
+/// a VM that lived a couple of minutes), NOT the cost of tearing it down. See
+/// [`process_cpu`] for the subtraction that turns this into a reclaim figure.
+fn reaped_children_cpu() -> std::time::Duration {
+    let mut usage = std::mem::MaybeUninit::<libc::rusage>::uninit();
+    // SAFETY: getrusage fully initializes the rusage it is given; we only read it on success.
+    if unsafe { libc::getrusage(libc::RUSAGE_CHILDREN, usage.as_mut_ptr()) } != 0 {
+        return std::time::Duration::ZERO;
+    }
+    let usage = unsafe { usage.assume_init() };
+    let to_duration = |t: libc::timeval| {
+        std::time::Duration::new(t.tv_sec.max(0) as u64, (t.tv_usec.max(0) as u32) * 1000)
+    };
+    to_duration(usage.ru_utime) + to_duration(usage.ru_stime)
+}
+
+/// CPU (`utime + stime`) consumed so far by a LIVE process, from `/proc/<pid>/stat`.
+///
+/// Sampled just before the SIGKILL, this is what a child had already spent while doing its
+/// job. Subtracting it from that child's whole-life CPU (credited to `RUSAGE_CHILDREN` when
+/// it is reaped) leaves exactly the CPU it accrued after the signal — i.e. its exit path,
+/// which for the VMM is dominated by `exit_mmap()` unmapping the guest's multi-GiB address
+/// space. Returns zero if the process is already gone, which biases the reported reclaim
+/// UP, never down: this number exists to stop teardown being called free.
+///
+/// `comm` (field 2) can contain spaces and parentheses, so fields are counted from after
+/// the last `") "`; `utime`/`stime` are fields 14/15, i.e. indices 11/12 from there.
+fn process_cpu(pid: u32) -> std::time::Duration {
+    let Ok(stat) = std::fs::read_to_string(format!("/proc/{}/stat", pid)) else {
+        return std::time::Duration::ZERO;
+    };
+    let Some(after_comm) = stat.rsplit_once(") ").map(|(_, rest)| rest) else {
+        return std::time::Duration::ZERO;
+    };
+    let fields: Vec<&str> = after_comm.split_whitespace().collect();
+    let ticks: u64 = [11usize, 12]
+        .iter()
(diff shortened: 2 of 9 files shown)


ejc3:
Verification against the final (post-amend) build.

**Reaping test, final code:**
```
$ make test-root FILTER=sigkill STREAM=1
Our tree: firecracker=2611012, holder=2611007, pasta=2611016
firecracker, holder and pasta all reaped by the kernel
test_sigkill_reaps_rootless_vm_tree PASSED
     Summary [5.135s] 1 test run: 1 passed, 687 skipped
```

**Corrected `reclaim_cpu_ms`,** on a VM whose container exited normally (the SIGKILL
test never reaches `cleanup_vm`, so it produces no teardown record):

```
vm teardown complete caller_blocking_ms=98 until_gone_ms=98 reclaim_cpu_ms=3 \
                     processes_us=174 network_us=25857 disk_us=72203
```

`reclaim_cpu_ms=3` where the pre-correction build reported `12328` for a comparable
teardown — that number was the VMM's whole-life CPU, credited to `RUSAGE_CHILDREN` at
reap. It now reflects only post-SIGKILL cost.

This record also shows the honest limit of the change. Here the guest had already shut
itself down, so the VMM was gone before teardown started (`processes_us=174`, i.e.
0.17ms) and there was no address-space reclaim for pasta's ~26ms exit
(`network_us=25857`) to hide under. Overlapping only helps when there is something to
overlap with; on the clone hot path — where firecracker is SIGKILLed holding a ~1GiB
mapping — that is exactly the case, and pasta's exit disappears underneath it. This is
why the log reports the phases separately instead of one "teardown was fast" number.

ejc3:
**Interaction with #731** (bench: CDP-direct request path) — both touch
`tests/test_signal_cleanup.rs`.

Trial-merged locally: **no conflict** (#731 appends a new test at end of file, this PR
edits the middle):

```
$ git merge --no-commit --no-ff origin/bench-cdp-request-path
Automatic merge went well; stopped before committing as requested
```

One follow-up for whichever merges **second**: #731's new
`test_bench_fast_teardown_leaks_nothing_clone` refers in its doc comment to
`test_sigkill_kills_firecracker_rootless`, which this PR renames to
`test_sigkill_reaps_rootless_vm_tree` (it now asserts the holder and pasta too, not just
firecracker). Comment-only, so it will not fail the build — it just needs the name
updated. I have deliberately not touched #731.

Worth noting the two PRs agree on the load-bearing point from opposite directions:
#731's test asserts the harness must reap on-disk state synchronously because
`cleanup_vm` never runs under SIGKILL, and this PR keeps every on-disk step inside
`cleanup_vm`'s blocking window for the same reason — no janitor, nothing deferred.

ejc3:
**Local `FILTER=snapshot` deferred to CI — box contention.**

Throughout this work the host had 2-4 concurrent `cargo nextest` VM suites from other
worktrees writing the same shared `FCVM_DATA_DIR=/mnt/fcvm-btrfs/root`. I started a
local snapshot run, then aborted it: a rebuild took the shared target lock mid-run and
swapped the binary under it, so its results were not trustworthy. I did not restart it
on a contended box, and I have stopped the queued retry rather than leave a VM suite
armed to fire unattended.

`Host-Root-{arm64,x64}-Snapshot{Enabled,Disabled}` in this PR's CI run the same suites
on dedicated runners and are the authoritative result. They are still queued.

Verified locally on the final code, all green:

- `make lint` (fmt + clippy `-D warnings` + audit + deny) — exit 0
- `make build`
- `make test-root FILTER=sigkill STREAM=1` — passed on every build, including after
  the review fixes
- teardown telemetry captured from real VM teardowns (see previous comment)

No stray VMs or processes were left behind from any of these runs.

ejc3:
## Tested: the sentence is **true** — but the reason given for it is **wrong**, and the correct reason carries a condition

You asked whether "this makes pasta's death depend on nothing but the kernel" is real or decorative, on the theory that `copy_process()` clears `pdeath_signal` in the forked child. I ran the discriminating test you described: hold the namespace-holder **alive** (so passt's PID watch cannot fire), then SIGKILL the process that armed the signal. Anything that kills pasta under those conditions is pdeathsig and nothing else.

| case | result |
|---|---|
| **positive control** — arm, `exec /bin/sleep` | died **3 ms** after parent SIGKILL (harness arms correctly) |
| **pasta, armed** — holder alive | **died 54 ms** after parent SIGKILL → **pdeathsig fired** |
| **pasta, unarmed** — holder alive | **survived 5 s** → without the prctl, nothing kills it |

The negative control is what makes this attributable: with the holder alive, unarmed pasta just sits there. Repeated against a holder userns created by root and one created by an unprivileged user: died in **3 ms** and **2 ms**. So the ten lines are load-bearing, not decorative — **keep them**.

**The fork hypothesis is disproven.** `--foreground` pasta does not fork: measured `threads=1`, and pasta's own `-P` pidfile reports **the same PID the pre_exec armed** (`657753`), with `ps` showing that PID parented directly to the harness. The process that persists *is* the process that was armed.

**But the comment's stated reason is incorrect.** It says the signal survives because "its namespace entry happens in a forked child". It doesn't happen in a forked child — there is no fork — and the credential change it's worried about **does** occur in the surviving process: measured `CapEff` goes `0x1ffffffffff` → `0x2014c2`, and pasta's userns is confirmed identical to the holder's. A credential change happened, in the armed process, and `pdeath_signal` survived it anyway.

The real reason is the subset test in `commit_creds()`:

```c
if (!uid_eq(old->euid, new->euid) || ... || !cred_cap_issubset(old, new)) {
        task->pdeath_signal = 0;
```

Under `sudo`, pasta's *prior* credentials are full root (`0x1ffffffffff`) and `--runas 0:0` leaves uid/gid at 0→0, so the new capability set is a strict **subset** of the old one and the clearing branch is never taken. The signal survives because pasta **loses** privilege on entry, not because a fork shielded it.

**Why this matters rather than being pedantry:** the correct reason has a precondition the stated reason does not. The subset argument holds only while fcvm runs as root. If fcvm ever runs as a genuinely unprivileged user, pasta's prior creds are that user's, entering the userns as root-in-userns is a capability **gain**, `cred_cap_issubset` fails, and the kernel zeroes the signal — silently returning pasta to the 1-second PID-watch path this comment says it no longer depends on. A comment that credits a nonexistent fork will not prompt anyone to re-check that.

Suggested: replace the "forked child" sentence with the subset rule and its precondition. Conclusion unchanged, code unchanged.

**Scope:** I did **not** test fcvm running as non-root — my second case varied the *holder's* userns owner, but the harness itself was still root, so both rows are the root/sudo shape. The non-root case is argued from the kernel rule above, not measured.

Reproducer: `pdeath_harness.c` (faithful fork→prctl→getppid→execv mirror of the shipped `pre_exec`) plus two driver scripts, in this session's scratchpad.

ejc3:
Applied in 3ee2e4f6 — comment only, no code change. Replaces the "namespace entry happens in a forked child" sentence with the `cred_cap_issubset` mechanism, and states the root-only precondition explicitly (plus the fact that it is argued from the kernel rule rather than observed, since the verifying harness was itself root).

Issue #627, opened 2026-06-07
Flaky: test_localhost_archive_mode fails intermittently under SnapshotEnabled load


ejc3:
`fcvm::test_localhost_image::test_localhost_archive_mode` FAILed (~58s) on `Host-Root-x64-SnapshotEnabled` (PR #620, which only edits test_signal_cleanup.rs); passes on main and on re-run — a load flake, not a regression.

**Root-cause direction:** archive mode = fc-agent `podman load`s the image from a block device at boot; under load this can be slow or race the health-wait. Capture the guest fc-agent log (did `podman load` finish? container start?) and the host health-monitor timeline; decide timeout-under-load (scale budget) vs. real intermittent failure.

**Parallel mitigation:** nextest retries. Tracking: #619.

ejc3:
Root-caused and fixed in PR #657. The framing in this issue (archive podman load racing a health-wait) wasn't the mechanism — the verified chain: under SnapshotEnabled, the VM's cache-ready handshake funnels into the global 10-permit snapshot semaphore BEFORE the pause; the 11th+ VM queues with the guest running and its fixed 30s cache-ack deadline ticking; on expiry the guest returns CacheResult::Failed and launches the container out of step with the host's pause, losing the marker output.

Fix: the host now emits 'cache-wait' keepalives every 5s while queued/snapshotting (and drains the guest's probe bytes so its close can't RST away the ack); the guest extends its deadline per keepalive (10-min absolute cap) and judges expiry only after draining queued data. Unit-tested on both sides, including the split-cache-ack-across-reads case.

ejc3:
Fixed by #657 (merged): cache-wait keepalive protocol (pinned timer; probe-byte draining; deadline extension with 10-min cap; expiry judged after drains). Passing across the full CI matrix including SnapshotEnabled.

PR #737, opened 2026-08-08
Replace three sleep-based race covers with real signals

ejc3:
Three sites from the codebase-wide sleep audit where a fixed delay was standing in for a signal that already existed. Each is a separate commit; they are grouped here because they are one concern — a sleep used as a race cover — and they touch disjoint code.

None of these is "the same sleep, but shorter": every one now waits on a real event, with a bound that only covers the case where the event can never arrive.

## 1. chronyd blocked every cold boot for ~1.5s — and failed silently

`fc-agent/src/agent.rs`, awaited inline from `run()`.

The audit flagged two sleeps (a 500ms "pkill worked" and a 1s "the command socket must be up by now"). Probing a live guest showed both were guarding code that could not work:

- The NTP servers were read from `/etc-host/chrony.conf`. **Nothing in the tree ever mounts `/etc-host`** — the string occurs only in that one function — so the list was always empty and the `chronyc add server` loop the 1s sleep existed for ran **zero times**.
- chronyd was started as `chronyd -u root` with **no `-f`**, so the config fc-agent wrote was read by nobody. chronyd loaded its compiled-in default, the rootfs's `/etc/chrony/chrony.conf`. The `makestep 1 -1` that lets a restored VM step a multi-hour clock jump was silently replaced by `makestep 1.0 3` — step only during the first three updates. Post-restore is exactly where chrony is supposed to be the correction.

Measured in a guest, before → after:

| | before | after |
|---|---|---|
| chronyd argv | `/usr/sbin/chronyd -u root` | `/usr/sbin/chronyd -f /etc/chrony.conf -u root` |
| makestep in force | `1.0 3` (rootfs config) | **`1 -1`** (fc-agent's, as intended) |
| `/etc-host` | does not exist | not used |
| `chronyc -n sources` | 4 (from the rootfs pool) | 4 (same pool, copied into fc-agent's config) |

Fixes: spawn the setup instead of awaiting it (the initial clock is already set synchronously by `sync_clock_from_host()` further up, so nothing in boot depends on chronyd); pass `-f` explicitly; build that config from the rootfs config's own `server`/`pool` directives so `rootfs-config.toml` stays the single source of truth for *which* servers a guest uses; replace the 500ms sleep with `systemctl stop chrony` plus a **pidfd** per leftover process (readable exactly when it exits, and `pidfd_send_signal` so a recycled PID can never be hit); replace the 1s sleep with a checked `chronyc` retry that ends on a real answer.

And it is no longer silent: the source count is logged on success, and an ERROR is logged when chronyd cannot be reached or reports zero sources.

Boot log after the change:

```
07:28:03.633 [fc-agent] output vsock connected (port 4997)
07:28:03.677 [fc-agent] chronyd started with 4 NTP source(s)
07:28:03.713 [fc-agent] root filesystem is btrfs, resized to fill disk
```

~44ms, concurrent with the rest of boot, where it previously blocked for at least the 1.5s of sleeps.

## 2. The cache-ack handshake closed on a 250ms grace window (#627's failure mode)

`src/commands/podman/listeners.rs` + `fc-agent/src/container.rs`.

The host wrote `cache-ack`, drained guest probe bytes for a fixed 250ms, then dropped the connection. A liveness probe arriving after that window was left unread — and an unread byte at close turns the close into a vsock RST that can flush the guest's receive buffer **before** it reads `cache-ack`, so the guest classifies a cold start as a warm restore. A window is a guess: a probe landing one millisecond late reopens the whole race.

Now: fc-agent closes the connection the moment it has read the ack, before it even logs, and the host drains **to EOF**. EOF is proof the ack landed and no probe is in flight, so the close cannot race the read.

The drain is **detached**, which matters: the next connection carries `ready`, and the documented post-resume vsock close-propagation failure (which made a second `read_line` on this task cost 300s) must not delay it.

## 3. Four copies of "sleep 200ms and hope the stderr reader drained"

`src/utils.rs`, `src/firecracker/vm.rs`, `src/hypervisor/cloud_hypervisor/mod.rs`, `src/network/pasta.rs` (2 sites).

`spawn_streaming` discarded both reader `JoinHandle`s, so a caller that saw its child exit early had nothing to await — and under load reported `no stderr captured` exactly when the cause mattered. It now returns them, and the exited-early paths await stderr EOF (2s bound, only for a grandchild holding the pipe open).

## Test Results

```
$ make lint

$ make test-root FILTER=sanity
    PASS [   7.937s] (1/6) fcvm::test_sanity test_sanity_rootless
    PASS [   8.192s] (2/6) fcvm::test_sanity test_sanity_bridged
    PASS [  17.849s] (3/6) fcvm::test_sanity test_sanity_routed
    PASS [  22.297s] (4/6) fcvm::test_fc_mock test_fc_mock_sanity
    PASS [  22.416s] (5/6) fcvm::test_sanity test_sanity_chrony_ntp_configured
    PASS [  28.071s] (6/6) fcvm::test_sanity test_ftrace_sanity
 Summary [  28.074s] 6 tests run: 6 passed, 683 skipped

 Summary [1251.705s] 116 tests run: 116 passed, 573 skipped

$ make test-root FILTER=chrony     # re-run against the final binary
    PASS [   6.744s] (1/1) fcvm::test_sanity test_sanity_chrony_ntp_configured
    ✅ chrony: 4 NTP source(s) from fc-agent's config
```

The new `test_sanity_chrony_ntp_configured` asserts all three links of the chain, because the failure mode is silence: every running chronyd uses `-f` on fc-agent's config, that config carries `makestep 1 -1`, and the daemon really reports at least one NTP source. The pre-fix guest runs `/usr/sbin/chronyd -u root`, which does not satisfy the first assertion — the test discriminates the bug rather than merely passing.

## Summary by CodeRabbit

* **New Features**
  * NTP synchronization now starts without delaying system boot and verifies that a time source is available.
  * NTP settings from the system configuration are preserved, with a default time source used when none is configured.

* **Bug Fixes**
  * Improved cache-handshake completion and connection cleanup.
  * Improved startup error reporting for virtualization and networking components by capturing complete process output.

* **Tests**
  * Added integration coverage verifying NTP configuration and readiness.

### DIFF
--- a/fc-agent/src/agent.rs
+++ b/fc-agent/src/agent.rs
@@ -161,10 +161,12 @@ pub async fn run() -> Result<()> {
     };
     let has_shared_volume = mounted_fuse_paths.iter().any(|p| p == "/mnt/shared");
 
-    // Start chronyd for ongoing NTP time sync.
-    // Must be after FUSE mounts so /etc-host/chrony.conf is readable.
-    // makestep 1 -1 allows stepping the clock at any time (critical after snapshot restore).
-    start_chronyd().await;
+    // Start chronyd for ongoing NTP time sync, OFF the boot critical path.
+    // Nothing below depends on the daemon being up, and the setup costs a
+    // process spawn plus a command-socket round trip — awaiting it inline made
+    // every cold boot wait for chrony before it could mount disks and launch
+    // the container. The task logs its own outcome, including the loud failure.
+    tokio::spawn(start_chronyd());
 
     let mounted_disk_paths = if !plan.extra_disks.is_empty() {
         eprintln!(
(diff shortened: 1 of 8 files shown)


ejc3:
Cross-check from the closed duplicate #734: its auto-fix PR #736 found a real gap that this branch should be checked against.

**`start_pasta()`'s deadline-timeout branch kills pasta and renders `stderr_tail_message()` WITHOUT first awaiting stderr EOF** (#734's src/network/pasta.rs:659-665). The kill closes the pipe's write end, which starts the EOF sequence, but the bail races it — so the timeout path can report "no stderr captured" precisely when a dead pasta's stderr is the thing you need.

Your branch adds the EOF await to the paths that report a dead pasta, but please confirm the DEADLINE-TIMEOUT branch specifically is covered, not just the spawn-failure and early-exit paths. If it is not, port it — the failure mode is identical to the one this PR exists to fix, and it only shows up under the load where you most need the diagnostic.

#734 is closed as a duplicate of this PR (I had two agents on the same slice — my error). Both of you independently found that /etc-host/chrony.conf is never mounted and that chronyd runs without -f, which is good corroboration. #736 is an auto-fix against the closed branch, so it will be closed too once this is confirmed.

ejc3:
Non-blocking note, recording a behavior change so it is not lost when #734 closes.

#734 was closed in favour of this PR. One functional difference between them survives, and it is a change **away from main**, not merely a feature #734 had and this does not.

`main`'s `fc-agent/src/agent.rs:517-530` reads the **host's** NTP servers out of `/etc-host/chrony.conf` (the FUSE-mounted host `/etc`) and injects them via `chronyc`. This PR instead reads the **guest rootfs** `/etc/chrony/chrony.conf` and falls back to `pool pool.ntp.org iburst`.

That substitution is coherent with the rest of the change and I am not asking for it to be reverted: moving chronyd off the boot critical path means it can no longer depend on the FUSE mount being up, and `/etc-host/chrony.conf` is exactly a FUSE-dependent read (main's own comment says "Must be after FUSE mounts so /etc-host/chrony.conf is readable"). Trading a FUSE-ordered read for a rootfs-local one is the right call for a task that must not be ordered.

What I checked rather than assumed, so the risk is stated accurately:

- **On this host it is benign.** `/etc/chrony/chrony.conf` here is `pool ntp.ubuntu.com` + the `*.ubuntu.pool.ntp.org` pools — functionally the same public pools as the fallback. Nothing changes.
- **On AWS it is plausibly an improvement, not a regression.** The usual host source there is the Amazon Time Sync Service at `169.254.169.123`, which is **link-local** and therefore not routable from a guest behind pasta/NAT. If that is what the CI runners carry, main has been copying an unreachable server into every guest, and the `pool.ntp.org` fallback is strictly better wherever the guest has egress. I did not verify a runner's chrony.conf, so this is stated as the likely case, not a measurement.
- **The genuine loss case is narrow:** a restricted-egress environment whose host chrony.conf names an internal, guest-routable NTP server. There, main works and this does not.

No action requested on this PR. If someone wants host propagation back later, the shape is to keep the read but do it inside the spawned task after the FUSE mount signal, rather than restoring the boot-path ordering this PR exists to remove.

ejc3:
Addendum to my note above, and an upgrade to it.

I filed the host-scoped NTP concern there as a hypothesis about AWS's link-local `169.254.169.123`, explicitly flagged as "the likely case, not a measurement." It is neither anymore. #734's branch `sleep-signals` carries `c98003e2 Never ship host-scoped NTP addresses into a guest` — the same problem, already root-caused and fixed, on the branch I closed as a duplicate of this one.

That does not change my dedupe call: this PR ran the full suites and #734 deferred them, which is the criterion I have applied all session. But it does mean #734 was not a strict subset of this PR, and I closed it without a commit-by-commit diff. That was the mistake, and it is mine.

Still no action requested here. This PR merges as-is. The orphaned commits — the host-scoped address filter, the `chrony.conf` change, and the `spawn_stderr_reader` helper that #736 depends on — get ported in one follow-up branch off main afterwards, re-verified rather than cherry-picked on faith. Details and the full list are on #736, which I have converted to draft because its base is the dead branch.

ejc3:
Correcting my two comments above. **I was wrong about the NTP behaviour change, and the error mattered.**

I claimed `main` read the host's NTP servers from `/etc-host/chrony.conf` and injected them via `chronyc`, and framed this PR as a change away from that — a possible regression in restricted-egress environments.

`/etc-host` **is never mounted.** In the pre-merge tree it appears in exactly three places, and all three are the same dead reference:

```
$ git grep -n "etc-host" origin/main~1
fc-agent/src/agent.rs:165:    // Must be after FUSE mounts so /etc-host/chrony.conf is readable.
fc-agent/src/agent.rs:514:/// Writes a chrony config using the host's NTP servers (from /etc-host/chrony.conf),
fc-agent/src/agent.rs:519:    let host_conf = std::path::Path::new("/etc-host/chrony.conf");
```

One comment, one doc comment, one path constant. Nothing creates it, nothing mounts it, no `--map` targets it. And the read was:

```rust
if let Ok(content) = tokio::fs::read_to_string(host_conf).await { ... }
```

`if let Ok` — the failure path is empty. The file never existed, the read always failed, `server_addrs` stayed empty, and `chronyc add server` was then called with **nothing to add**. Every guest got zero NTP sources, silently, forever.

So this PR did not regress host propagation. There was no host propagation. It replaced a silent no-op with a real, reachable source, which is strictly better. **My "genuine loss case" paragraph describes a capability that never existed.**

Credit where due: this was found by the agent on #734 — the PR I closed as a duplicate of this one — which is also why I am not treating that branch as fully superseded. Its remaining unique work is real:

- resolve the host's NTP servers **host-side into the boot plan**, which is how you actually get them into a guest without a mount that does not exist, with host-scoped (loopback/link-local) addresses filtered out because they name nothing inside a guest;
- put the sources in `chrony.conf` instead of adding them at runtime. The comment justifying `chronyc add server` claimed chrony cannot parse bare IPv6 literals. That claim is **false** — verified on chrony 4.5, a config whose only source line is `server 2620:2d:4000:1::3f iburst` drives `chronyd -Q` to a real measurement and exit 0. Removing that workaround deletes the whole command-socket dependency.

This is the same failure mode this repo keeps surfacing: a code path that reads as working, cannot possibly work, and is silent about it. I reproduced the error by reading the code and believing it, rather than checking whether the file it names is ever created.

No action on this PR — it is merged and it is an improvement. The correction is on the record so the follow-up is scoped against what is actually true.

PR #740, opened 2026-08-08
Make the UFFD memory server fail closed and bound its blast radius

ejc3:
Makes the UFFD memory server **fail closed** and bounds its blast radius. Addresses three CRITICAL findings from today's codex review.

## The Problem

**A closed userfaultfd does not stop a guest.** The kernel resolves every subsequent fault with a **zero page**. So when a clone's page-fault handler died, `server.rs:234-244` logged that the VM "should be killed" — and left it running, executing against memory that was silently wrong, free to commit that wrongness to its disk, to a volume, or to a response it returned to a caller. `server.rs:282-286` documented this as silent memory corruption. For a request service, a corrupted answer is the worst possible outcome; there was no authoritative failure signal anywhere.

Two more, same file:
- **Unbounded shared failure/fairness domain.** One server accepted any number of clones, each spawning a task that drained *every* available fault before yielding. One clone could hold a runtime worker thread and starve the rest.
- **Socket path race.** The server blind-unlinked a deterministic path on startup *and* on drop, so overlapping servers unlinked each other's sockets.

### 1. Fail closed — kill the clone whose memory cannot be served

There is no in-band way to tell a restored Firecracker "your memory backend is gone", so the server stops the guest before it can act on zeros.

New `ClonePeer` identifies the connecting VMM from **`SO_PEERCRED`** on the server's own socket — the kernel's answer to "who opened this connection", not anything the peer claims — pinned by `(pid, start_time)`, the same process identity fcvm's state files use, so a reused PID is never signalled. `kill_fail_closed()` SIGKILLs it (not SIGTERM: a graceful exit would run guest/VMM code over unserved pages) and logs `FAIL CLOSED`.

Wired to **every** path that ends a clone's service: handler error, handshake failure, handshake timeout, clone-cap refusal, and — via a task-id→peer map reaped in `UffdServer::reap` — a **panicking** handler, which runs none of its own error path but closes the uffd just the same. A connection whose VMM cannot be identified is refused rather than served: a clone this server could not fail closed on must not be taken on.

**The signal the caller sees.** In `cmd_snapshot_run`, `abnormal_vmm_exit` classifies a VMM that was signal-killed or exited non-zero, marks the clone `VmStatus::Failed` / `HealthStatus::Unhealthy`, and makes the command **exit non-zero**. Previously a killed clone returned exit 0 — a corrupt run reported as a good one.

### 2. Bound the blast radius

- **`FCVM_UFFD_MAX_CLONES`** (default 64) caps clones per server. Authoritative at accept (after reaping finished handlers, so the count is exact and no clone is refused spuriously); advisory in `snapshot run`, which counts live clones on the serve PID and errors *before* spawning a VM — same two-layer shape as the existing `preflight_clone_hugepages`. A malformed or zero value is refused, not silently ignored.
- **Fairness:** `drain_events` serves at most **128** faults, then returns `DrainOutcome::BatchLimit` and the handler yields. Readiness is deliberately **not** cleared on that path — `AsyncFd` is edge-triggered, and clearing it would park the handler waiting for an edge that cannot come while the guest is blocked on those very faults.

### 3. Socket path: unique per instance, never unlink a live one

`new_with_path` (caller-chosen path) and the deterministic `new` are **gone**. One constructor derives its own name: `uffd-{snapshot}-{pid}-{pid_start_time}.sock` — reusing fcvm's existing PID+start-time identity as codex suggested, not a new mechanism. Two live servers cannot derive the same name, which is what makes unlinking safe at all.

Startup no longer blind-unlinks: `claim_socket_path` probes first — a **live** listener means refuse to start; only a proven-dead socket is cleared. `Drop` unlinks only a socket this server actually bound. The serve command's own `remove_file` is deleted (it raced the server's `Drop` and could unlink a still-listening server's socket).

The path is **published** in `VmConfig::uffd_socket` and read back by clones and by `poll_serve_ready`, replacing two independent re-derivations of the same string. Over-long paths are refused by name instead of a bare `EINVAL` out of `bind(2)`.

Also: `process_start_time` moved from `state::manager` to `utils`, beside its siblings `is_process_alive` / `is_same_process_name` — one definition, in the module whose callers signal based on the answer.

### Test instrumentation

`FCVM_UFFD_FAIL_AFTER_FAULTS=<n>` fails a handler after `n` faults, checked once per batch (unarmed = one relaxed load per ≤128 faults). Same `FCVM_*` env-gated style as `FCVM_FUSE_TRACE_RATE` and the `failpoint` crate. It exists because there is no other honest way to exercise the path: what must be guaranteed is what happens to a **healthy, running** guest the moment its server fails, and a real Firecracker never makes its own memory server fail.

## Test Results

```
$ make lint
advisories ok, bans ok, licenses ok, sources ok      # fmt + clippy -D warnings + deny

$ make build
Finished `release` profile [optimized] target(s)

$ make test-unit
Summary [50.816s] 409 tests run: 409 passed, 0 skipped
```

7 new unit tests in `uffd::server`, all passing:

```
PASS test_kill_fail_closed_kills_the_clone_vmm       # real process, reaped with SIGKILL
PASS test_kill_fail_closed_refuses_a_reused_pid      # mismatched start_time => survives
PASS test_clone_peer_identity_comes_from_the_kernel  # SO_PEERCRED, not self-reported
PASS test_claim_socket_path_refuses_a_live_listener  # live socket survives
PASS test_claim_socket_path_clears_a_dead_socket
PASS test_socket_path_is_unique_per_server_instance
PASS test_socket_path_rejects_oversized_paths
PASS test_max_clones_parsing                         # 0 and garbage refused
```

`test_kill_fail_closed_kills_the_clone_vmm` is the regression test for the core bug: it spawns a real process and asserts it is actually **reaped with SIGKILL**. The old code only logged.

New integration test `tests/test_uffd_fail_closed.rs::test_uffd_handler_failure_kills_and_fails_the_clone` arms the handler failure on a real serve process, then asserts the clone **exits** (bounded wait — *still running* is precisely the corruption mode), **exits non-zero**, is gone from the process table, that the serve log records `FAIL CLOSED`, and that the **serve process survives** the clone it killed (blast radius bounded).

## Not yet run locally

The VM lanes — `make test-root FILTER=snapshot` and the new integration test — were **deferred**: the box had three concurrent nextest suites and load 38.75. Testing VMs under that contention is how four datasets were discarded today. CI runs both lanes on this PR.


ejc3:
Cross-posting from the author of **#739**, the duplicate implementation of this task, with the evidence needed to pick between the two.

**First, a correction on attribution.** #739 is my branch (`agent/uffd-fail-closed-svc`); #740 is `uffd-fail-closed`. I was asked whether "my SO_PEERCRED approach" closes the PID-reuse window — but `SO_PEERCRED` + `pidfd_open` is precisely the *buggy* pattern, and it is the one I **removed**. Both PRs started there. Answering the two questions honestly:

## 1. The PID-reuse window

**#739 had it, and it is now closed.** Codex flagged it as P1 on my first commit — exactly as described: two syscalls (`process_start_time` then `pidfd_open`) with a reap-and-recycle window between them, after which the pidfd pins a stranger. Fixed in `c7dca887` by taking the identity from the socket in **one atomic step**:

```rust
// SO_PEERPIDFD (Linux 6.5+) — a pidfd for the peer recorded at connect time.
let pidfd = peer_pidfd(stream.as_raw_fd())?;
```

Nothing is addressed by PID afterwards: the PID is read back *out* of the pinned handle for logs only, and `is_alive()` asks the pidfd via `pidfd_send_signal(fd, 0)`. Kernel support is probed once at server construction, so an unsupporting kernel is a loud startup failure rather than a per-connection surprise. **PID recycling cannot misdirect it, because no decision uses a PID.**

**#740 still has a window, narrower but real.** `ClonePeer` carries `{ pid, start_time }` with no pidfd anywhere, and `kill_fail_closed` does:

```rust
let current = crate::utils::process_start_time(self.pid);   // check
if current != self.start_time { return; }                   // ...
nix::sys::signal::kill(Pid::from_raw(pid), SIGKILL);        // act
```

That is a check-then-act TOCTOU. Between the comparison and the `kill(2)`, the peer can exit, be reaped, and have its PID recycled — and the SIGKILL lands on the stranger. Credit where due: this recheck is a genuine mitigation my first version lacked, and it shrinks the window from *connect-to-kill* down to *microseconds*. But "narrow" is not "zero", and the repo's rule is ZERO RACES. A pidfd removes the window by construction rather than shrinking it; that is the entire reason the primitive exists.

## 2. The fail-open accept path

**Both PRs have it; #740 is much harder to trigger.** #740's `from_stream` failure does a bare `continue`, dropping the accepted stream. But its `from_stream` only calls `peer_cred()`, which on a connected `AF_UNIX` socket essentially cannot fail — whereas mine also called `pidfd_open`, giving it a reachable `EMFILE` path. So this is a live bug in mine and a latent one in #740.

I fixed it in `c7dca887` by never dropping an unpinnable peer (`kill_unpinned_peer`, via `SO_PEERCRED` since it needs no fd). #740 would want the same guard even though reaching it is unlikely.

## 3. The finding that decides it: **`DEFAULT_MAX_CLONES_PER_SERVER = 64` is broken, and #740 ships it**

This is the one I would act on first. #740 line 33 has the same `= 64` I shipped originally. I ran the suite against it:

```
$ make test-root FILTER=snapshot STREAM=1        # cap=64
     Summary [ 481s] 114 tests run: 112 passed, 2 failed
        FAIL fcvm::test_snapshot_clone test_snapshot_clone_stress_100_bridged
        FAIL fcvm::test_snapshot_clone test_snapshot_clone_stress_100_rootless

  ERROR uffd: REFUSING clone: ... active_clones=64 max_clones=64
  ERROR uffd: KILLED the clone's VMM ... reason=server is at its concurrent-clone cap
  ERROR fcvm: Error: loading snapshot: connection closed before message completed
```

`test_snapshot_clone_stress_100_*` restore **100 clones from one serve process**. At 64 the server refuses and kills clones 65..100 — working exactly as designed, on a configuration fcvm supports. Nothing else in the suite failed; it was the cap and only the cap.

Raised to 256 in `249e7264`, re-run clean:

```
        PASS [  22.785s] (1/2) test_snapshot_clone_stress_100_bridged
        PASS [  20.706s] (2/2) test_snapshot_clone_stress_100_rootless
     Summary [  43.494s] 2 tests run: 2 passed, 692 skipped
     cap refusals: 0
```

And the floor is now a build error, not a test failure (`785b81bd`), verified by breaking it:

```
error[E0080]: evaluation panicked: the default UFFD clone cap must exceed the
              per-server fan-out the stress tests restore
```

**#740 should expect both stress tests to fail in CI for this reason.**

## 4. A contradiction between the two PRs worth settling

We document *opposite* consequences for dropping the connection. #740:

> in COPY mode Firecracker never waits for a reply, so it would restore and then **hang forever** on its first fault

#739: the in-flight uffd is released and the guest runs on **zero-filled pages** (silent corruption). Both PRs use this claim to justify the same design, and they cannot both be right — a hang is loud and safe, silent zeros are the thing worth killing a VM to prevent. Whichever PR survives, this should be settled by test rather than by comment, because it determines whether the fail-open path is a corruption bug or merely a hang bug.

---

Not arguing for my branch over this one — #740 has real strengths (the already-exited `Option<u64>` handling is careful, and its per-clone `vm_id` in the peer makes the logs better than mine). But on the specific questions asked: **the pidfd window is closed in #739 and open in #740, and the `= 64` cap is measured-broken in both.** Happy to hand the `SO_PEERPIDFD` helper and the cap fix to whoever's branch survives.

ejc3:
Closing in favour of #739 — same task, two agents, my error again (fourth duplicate today).

Picking #739 on measured evidence, not preference:

**1. Clone cap.** This PR ships `DEFAULT_MAX_CLONES_PER_SERVER = 64`. `test_snapshot_clone_stress_100_*` restores 100 clones from one serve process, so the server refuses and kills clones 65..100. Measured on the other branch: `114 tests run: 112 passed, 2 failed` — the cap and only the cap. At 256 both pass with zero refusals. #739 ships 256 plus a compile-time assertion (`DEFAULT_MAX_CLONES_PER_SERVER > EXERCISED_MAX_CLONES_PER_SERVER`) verified by negative control, so the regression cannot recur silently.

**2. PID-reuse window.** This PR re-reads `start_time`, compares, then calls `kill(Pid::from_raw(self.pid), SIGKILL)` — check-then-act. The recheck is a real mitigation (#739's first version lacked it) and shrinks the window to microseconds, but narrow is not zero, and this code SIGKILLs processes. #739 replaced the two-syscall `SO_PEERCRED` + `pidfd_open` pattern with `SO_PEERPIDFD` — one atomic step from socket to pidfd, so no kill decision is addressed by PID at all, with kernel support probed once at startup.

Where this PR is BETTER, and #739 should adopt it: your `from_stream` only calls `peer_cred()`, which on a connected AF_UNIX socket essentially cannot fail, so the fail-open accept path is latent here rather than live. #739 had a reachable `EMFILE` path there (now fixed).

**One thing that must be settled rather than lost:** the two PRs document OPPOSITE consequences for dropping the connection — #739 says the guest gets zero-filled pages (silent corruption), this one says it hangs forever. Both use the claim to justify the same design, so one of them is wrong. That difference decides whether the path is a corruption bug or a hang bug, and it should be settled by test on #739 before merge.

Nothing here is lost: the work is equivalent, the cap and pidfd differences are the deciding factors, and the fail-open observation is carried across.

PR #742, opened 2026-08-08
Replay a snapshot's recorded working set instead of faulting it in

ejc3:
Replay a snapshot's recorded guest working set on restore, so a UFFD clone stops
paying a fault for pages we already know it will touch.

## The Problem

A UFFD-restored clone starts with nothing resident and takes a vCPU trap plus an
ioctl for every page it touches. Measured on a Chromium snapshot
(`bench/chromium`, 2026-08-08, clean exclusive box): **56,300 faults at ~5.6 us
marginal, +316 ms of caller latency** versus a file-backed restore of the same
image.

Readahead is the obvious fix and it does not work here. The measurement says why:

| Property | Measured | Consequence |
|---|---|---|
| next fault is the next page | **8.6%** | fault-around / sequential readahead guesses wrong ~91% of the time |
| faults by 300 ms / 750 ms | 50% / 98% | not front-loaded — warming an early burst misses most of the set |
| pairwise page-set Jaccard, 8 clones | median **0.927** | the SET repeats across clones |
| union faulted by ALL 8 clones | **82.2%** | ...and repeats almost exactly |

The order is unpredictable; the set is not. So persist the **set**.

## The Solution

**Record.** Every clone's faults are marked in a bitmap over 4 KiB granules of
the memory image. Fixed granularity rather than the clone's page size, so one
recorded set serves 4 KiB, 16 KiB and 2 MiB clones; it dedupes repeat faults for
free and merges two clones' observations with a word-wise OR.

**Persist.** The union lives beside the image as `memory.bin.working-set`. The
read-modify-write runs under an exclusive `flock` on a sidecar lock file and is
published by atomic rename to a uuid-unique temp name (not pid — nested VMs reuse
pids, see AGENTS.md). It only writes when the union grew, so the steady state
writes nothing. Keyed by SHA-256 of the image's `(len, mtime, ino, dev)`;
rewriting a snapshot writes a new file, so the stale set is dropped.

**Replay.** At the handshake — the earliest the uffd exists, and before any vCPU
runs (Firecracker is still loading the vmstate; fcvm only sends `Resumed` once
that returns) — the union's contiguous runs are bulk-populated on a blocking
thread, one `UFFDIO_COPY` (or `UFFDIO_CONTINUE` in MINOR mode) per 2 MiB chunk
instead of one per page. Runs are first cut to the regions THIS clone mapped and
to whole granules of ITS page size.

**Degrade.** Missing, stale, truncated or garbled ⇒ on-demand faulting. Never an
error, never a hang.

### Why a stale set cannot corrupt a guest

The set only says WHICH offsets to populate. The bytes always come from the image
being served right now, at that same offset, through the same code path a demand
fault uses. A wrong set therefore populates pages nobody needed (wasted copies)
or misses pages that then fault as before — it cannot place wrong bytes anywhere.
The key exists to keep replay USEFUL, not to keep it SAFE.

That is also why the key is a `stat` tuple and not a content hash: hashing the
image measured 1.5 GB/s on this host (1.4 s for a 2 GiB snapshot, on every
`snapshot serve` startup), far more than the mis-prefetch it would prevent.

### Isolation

Replay touches pages the guest never asked for, so it must not be visible across
clones — a previous `MAP_SHARED` experiment corrupted a golden snapshot. Under
COPY each clone gets a private page. Under MINOR replay installs a **read-only**
PTE onto a folio shared by every clone and the first guest write takes an
ordinary CoW fault. Both are proven by test, below.

## Test Results

35 unit tests in `src/uffd`. Three are end-to-end: they drive the real handshake,
the real replay ioctls and the real fault handler against a real userfaultfd,
with stand-in clones that map memory and register a uffd exactly as a restoring
Firecracker does — no VM required, fully deterministic.

```
$ cargo test --release --lib uffd::
test uffd::server::tests::working_set_replay_serves_pages_without_faults_and_keeps_clones_isolated ... ok
test uffd::server::tests::working_set_replay_on_shared_backing_still_isolates_clones ... ok
test uffd::server::tests::a_corrupt_working_set_falls_back_to_on_demand_faulting ... ok
...
test result: ok. 35 passed; 0 failed; 0 ignored; 0 measured; 216 filtered out
```

- `working_set_replay_serves_pages_without_faults_and_keeps_clones_isolated`
  3 MiB recorded of a 4 MiB image, so replay spans several ioctls per run.
  **Fault count: 0** for the 768 replayed pages in both clones, with the right
  content; the one unrecorded page faults **exactly once** and is served
  correctly; a write by one clone is invisible to the other; the image is
  byte-identical afterwards.
- `working_set_replay_on_shared_backing_still_isolates_clones`
  The same on the MINOR backing, where replay installs PTEs onto SHARED folios:
  the write CoWs, and the other clone *and the shared backing* both still read
  the pristine page.
- `a_corrupt_working_set_falls_back_to_on_demand_faulting`
  A garbage list replays nothing and all 1024 pages are served on demand with the
  right content. **Fault count: 1024** — the "before" number, in a test.

Mutation-checked that those tests actually bite:

| Mutation | Result |
|---|---|
| replay installs nothing | FAILS — `timed out waiting for first clone's replay` |
| chunk placed one page off | FAILS — `replayed page 1 has the wrong content: left 128 right 129` |
| clone maps the shared backing `MAP_SHARED` | cannot even mmap it (`EACCES`) — the seal from #721 |

```
$ make lint     # fmt + clippy --all-targets -D warnings + audit + deny
pass
$ make build
pass
$ make test-unit
Summary [50.492s] 422 tests run: 422 passed, 0 skipped
```

## Deferred

`make test-root FILTER=snapshot` and the before/after latency check on the server
path are **not run**. Two other agents were running the snapshot VM suite on this
box (load average 9-22) while this was finished; VM-testing or measuring under
that contention produces failures and numbers that mean nothing. CI will run the
snapshot suite on a dedicated runner.

## Knobs

- `FCVM_UFFD_WORKING_SET=off` — disables record and replay (the control arm for a
  before/after measurement). An unrecognised value is a hard error.
- `FCVM_UFFD_FAULT_TRACE=<dir>` — per-fault trace for measurement tooling:
  `<pid>-<vm_id>.faults`, little-endian `(image_offset, ns_before, ns_after)` u64
  triples. It shares the single per-fault observation point with the recorder, so
  the hot loop has one instrumentation path, not two.
- `fcvm snapshot serve` prints faults served and bytes replayed at shutdown.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1516,6 +1516,41 @@ fcvm snapshot run --pid <serve_pid> --name clone1
   clean pages via the page cache (MAP_PRIVATE) — measured 3x 1GiB clones
   ≈ 230MiB total PSS, with or without dirty tracking (#632)
 
+#### Working-Set Replay (`src/uffd/working_set.rs`)
+
+Laziness costs latency: every page the guest touches is a vCPU trap plus an
+ioctl (~5.6 us marginal, measured), and a Chromium clone takes ~56,300 of them —
++316 ms versus a file-backed restore of the same snapshot.
+
+**Why replay and not readahead.** Across 8 clones of one snapshot the pairwise
+page-set Jaccard median is 0.927 and 82.2% of the union is faulted by ALL 8, so
+the SET is predictable. The ORDER is not: only 8.6% of faults land on the page
+after the previous one, and faults are spread over the whole restore (50% by
+300 ms, 98% by 750 ms), so fault-around/readahead guesses wrong and warming an
+early burst misses most of the set.
+
+- **Record**: every clone's faults are marked in a bitmap over 4 KiB granules of
+  the memory image; on handler exit it is OR-ed into the union stored beside the
+  image as `memory.bin.working-set` (exclusive `flock` + atomic rename).
+- **Replay**: at handshake — before Firecracker finishes loading, so before any
+  vCPU runs — the union's contiguous runs are bulk-populated with one
+  `UFFDIO_COPY` (or `UFFDIO_CONTINUE` in MINOR mode) per run.
+- **Key**: SHA-256 of the image's `(len, mtime, ino, dev)`. Rewriting a snapshot
+  writes a new file, so the stale set is dropped.
+- **A wrong set cannot corrupt a guest.** It only says WHICH offsets to
+  populate; the bytes always come from the image being served, at that offset,
+  through the same path a demand fault uses. Missing/stale/corrupt ⇒ on-demand
+  faulting, never an error and never a hang.
+- **Isolation**: replay touches pages the guest never asked for, so it must not
+  be visible across clones. COPY gives each clone a private page; MINOR installs
+  a read-only PTE onto the shared folio and the first guest write CoWs.
+  `working_set_replay_*` in `src/uffd/server.rs` proves both, plus that the
+  snapshot image stays byte-identical.
+- **A/B switch**: `FCVM_UFFD_WORKING_SET=off` disables record and replay (an
+  unrecognised value is a hard error). `FCVM_UFFD_FAULT_TRACE=<dir>` writes a
+  per-fault trace (`<pid>-<vm_id>.faults`, LE `(offset, ns_before, ns_after)`
+  u64 triples) for measurement tooling.
+
 ### FUSE Parallelism (fuse-pipe)
 
 **Kernel clone fd model (`FUSE_DEV_IOC_CLONE`):**
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1534,7 +1534,9 @@ early burst misses most of the set.
   image as `memory.bin.working-set` (exclusive `flock` + atomic rename).
 - **Replay**: at handshake — before Firecracker finishes loading, so before any
   vCPU runs — the union's contiguous runs are bulk-populated with one
-  `UFFDIO_COPY` (or `UFFDIO_CONTINUE` in MINOR mode) per run.
+  `UFFDIO_COPY` (or `UFFDIO_CONTINUE` in MINOR mode) per 2 MiB chunk instead of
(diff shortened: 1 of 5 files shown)


ejc3:
Closing in favour of #744 — same feature, two agents, my error (fifth duplicate today; the branch left in the abandoned worktree is this one).

Deciding on evidence:

**#744 has the end-to-end proof, this PR defers it.** #744 measured `20,048 -> 3,585` demand faults (82% fewer) across two clones with identical lifecycles, and `make test-root FILTER=working_set_replay` PASSED (20.9s) including live cross-clone isolation and a byte-identical `memory.bin` check. This PR deferred both `make test-root FILTER=snapshot` and the before/after latency check — which is the measurement that proves the feature does anything.

**#744 also fixed a prerequisite bug this PR does not mention.** Handlers never exited: a userfaultfd gives NO signal when its creating process dies, because the server holds its own reference so `userfaultfd_release` never runs. Proven with a standalone C probe (poll returns 0 with revents=0 forever; read gives EAGAIN; only UFFDIO_COPY reveals it with ESRCH) and corroborated against history — `"VM exited"` appears **0 times across 137 serve logs**, so that branch was dead code and every finished clone pinned a task and a uffd until the server stopped.

**Your analysis is identical where it matters** — same 8.6%-sequential finding, same conclusion that readahead guesses wrong ~91% of the time, same design consequence. Independent agreement on that is worth something.

**Please port one thing across to #744:** `a_corrupt_working_set_falls_back_to_on_demand_faulting`. #744 tests EEXIST-during-replay and address-underflow planning, but a deliberately garbled list falling back to on-demand — and serving all pages correctly — is the failure mode most likely to hit in production, since the invalidation key there is (len, mtime, ino, dev) rather than a content hash. That key choice is well justified (SHA-256 of 2 GiB measured 1.4s per `snapshot serve`, and a stale set can only waste a copy, never corrupt), but it makes the corrupt-input test more valuable, not less.

Issue #661, opened 2026-06-10
test_clone_port_forward_stress_rootless: first clone's pre-storm curl RSTs (health-ready vs proxy-path-ready race; reproduces on main)


ejc3:
## Symptom

`test_clone_port_forward_stress_rootless` fails its **pre-storm verification**: the first-spawned clone's curl gets `curl: (56) Recv failure: Connection reset by peer` on the loopback-proxy path, while the other clones pass. Reproduces ~2/3 locally.

## Key facts (2026-06-10, Graviton3 dev host)

- **Reproduces on `origin/main`** (2/2 fails, same signature) on the same host — *not* introduced by the #630 branch. CI on main is green, so it is host-/timing-conditional.
- All three non-stress siblings (`test_clone_port_forward_{bridged,rootless,routed}`) pass consistently on the same host.
- Timeline from one failure: the curl raced the clone's restore revival — the guest's `restore complete (exec + egress + output reconnected)` landed ~0.3s after restore start, `pasta` "port forward ready" probe at +0.245s, and the failing curl hit inside that window.
- `poll_health_by_pid` gates the curl, but the health monitor probes the guest **directly** (nsenter → guest IP) while the test curls the **loopback proxy → pasta → guest** path — health-ready ≠ proxy-path-ready.
- "Reset by peer" (established, then RST) rather than "refused" suggests either the proxy's backend connect failing mid-flight, or the guest's `kill_stale_tcp_connections` (`handle_clone_restore`) murdering the in-flight test connection if the epoch watcher fires late.

## Suspects, in order

1. Health flips healthy before `verify_port_forwarding`'s probe chain is actually equivalent to the test's path (the verify runs before the health monitor spawns in `cmd_snapshot_run`, but verify-pass ≠ sustained-ready under multi-clone contention).
2. `kill_stale_tcp_connections` on a late restore-epoch trigger killing fresh inbound connections (check its filter against post-restore inbound flows).
3. Multi-clone restore contention (3 concurrent restores of the same snapshot) skewing the first clone's revival timing — the test recently dropped its retry workarounds (96040729), narrowing tolerance.

## Suggested approach

Instrument with the existing `FCVM_FORCE_DIAG` machinery on a loaded host; correlate the failing curl's timestamps with the guest's restore-handler phases; then either make health publication wait for the full revival (the principled fix) or tighten `kill_stale_tcp_connections`' filter.

[...]

PR #746, opened 2026-08-08
Tear down VMs concurrently, and measure what teardown really costs

ejc3:
Makes VM teardown concurrent instead of serial, and reports what it actually costs. Also re-diagnoses the "pasta has no PDEATHSIG" finding: the prescribed fix is a no-op, and the evidence is below.

## The Problem

~154ms of every benchmark request happened *after* the result existed. `src/commands/common.rs::cleanup_vm` killed-and-waited each child in turn:

```
SIGKILL firecracker -> wait  (~40ms: kernel reclaiming a ~1GiB guest mapping)
SIGKILL holder      -> wait
network.cleanup()   -> SIGKILL pasta -> wait  (~28ms)
on-disk reaping                              (~10ms)
```

The kernel unmaps the guest's address space in `exit_mmap()`, in the dying VMM's own context, charged to its system time. Everything else queued behind that reclaim while the machine was otherwise idle.

## The Solution

Five phases, with one load-bearing rule: **signal every child before waiting on any**.

1. Stop in-process tasks (health monitor, output listener, volume servers)
2. `start_kill()` the VMM, the holder and the network helper — **no `.await` between the signals**
3. `tokio::join!` their exits
4. Network cleanup (namespace/veth/iptables; reaps the already-dead helper)
5. On-disk reaping

Two-phase kill split out of the single `kill()`: `Hypervisor` gains `start_kill()`/`reap()` (with `kill()` now their composition, still used by error paths), and `NetworkManager` gains `start_kill_processes()` — default no-op, implemented by `PastaNetwork`. `cleanup()` still kills and reaps unconditionally, so calling it is optional and calling it twice is harmless.

**On-disk reaping stays synchronous.** State file, VM data dir, FC log, NFS exports. "Leave nothing on disk" is a correctness contract of the same rank as "leave no orphaned VM" — a janitor or a detached task would let the process that knows the paths exit before the work is proven done. Measured `disk_reap_ms` is 54-72ms; it is not where the time goes.

## Teardown is measured, not claimed

Every teardown logs three separate numbers, so it can never be reported as free:

```
VM teardown complete vm_id=… caller_blocking_ms=124.3 processes_gone_ms=61.0
                             reclaim_cpu_ms=58.8 net_cleanup_ms=0.15 disk_reap_ms=63.1
```

- `caller_blocking_ms` — what a request actually pays
- `processes_gone_ms` — SIGKILL → last child reaped. `exit_mmap()` runs before a task is reapable, so this returning *is* "truly gone"
- `reclaim_cpu_ms` — CPU the children burned dying

The CPU number needed care. `getrusage(RUSAGE_CHILDREN)` folds in a child's **whole-life** CPU when it is reaped, so a naive window delta read **~4950ms** — the VM's entire runtime, not its death. Subtracting each child's pre-kill `/proc` CPU leaves only the dying cost: **46-60ms**, tracking `processes_gone_ms`. The reclaim is genuine CPU work; overlapping teardown moves that cost, it does not delete it.

## pasta cannot use PDEATHSIG — evidence, not opinion

The review finding ("`src/network/pasta.rs` has no PDEATHSIG") is factually correct, but **adding one does nothing**, so this PR documents the mechanism instead of shipping a placebo.

passt changes its own credentials *after* exec. Given a bare PID, `conf_pasta_ns()` derives `/proc/PID/ns/user`, and `isolate_user()` does `setns(CLONE_NEWUSER)` into the holder's user namespace. That grants capabilities — measured on this host, pasta `CapEff 0x200400` (`CAP_SYS_ADMIN|CAP_NET_BIND_SERVICE`) versus fcvm's `0x0` — and `commit_creds()` zeroes `task->pdeath_signal` on any credential change that is not a capability subset. It is the same kernel rule that already forces `install_namespace_pre_exec`'s hook to run LAST, except here the transition happens inside passt, after exec, where fcvm cannot re-arm it.

Measured with the pre_exec hook installed:

| build | holder death | pasta death |
|---|---|---|
| pdeathsig hook **absent** | 2ms | 188ms |
| pdeathsig hook **present** | 2ms | 169ms / 699ms |

Unchanged — pasta stays on its own schedule. (An experiment routing around it via `--netns` was also tried and reverted: it does not avoid the userns transition either.)

**What does reap pasta:** passt's own `pasta_netns_quit_*` watchdog. Its target `/proc/<holder>/ns/net` vanishes when the holder dies (kernel-enforced, 2ms), and passt exits when it notices — but the path is on procfs, so passt cannot use inotify and falls back to `pasta_netns_quit_timer()`, a **one-second interval poll**.

**Residual risk, stated plainly:** pasta is never permanently orphaned, but it can outlive its VM by up to ~1s while still holding that VM's forwarded host-side listening sockets. With teardown now ~60ms, the next VM can be handed the dead VM's loopback IP inside that window — the routing hazard the review raised, now quantified. Closing it properly means patching passt to re-arm pdeathsig after its isolation completes (`pasta/patches/`, the same mechanism used for the #661 addr_seen fix). That is a vendored-fork + pinned-build change and is deliberately **not** bundled here.

## The leak guarantee is intact

No change to how any child is reaped on SIGKILL. Additionally hardened: one shared `utils::set_pdeathsig_sigkill(expected_ppid)` replaces two copies of the raw `prctl`, and re-checks `getppid()` after arming. That closes the fork/prctl race — a parent that dies in that window has already lost its parent-death notification and the child would have run forever; now the `exec` is aborted instead.

The existing reaping test was **extended, not duplicated**: `test_sigkill_kills_firecracker_rootless` now requires a pasta process to exist and asserts it is gone, proving **no permanent orphan**. It deliberately does *not* assert a tight deadline — passt's one-second poll would make that a coin flip — and the comment states exactly what is and is not proven, plus the instruction to tighten it if the passt patch lands.

## Test Results

```
$ make lint
cargo fmt -p fcvm -p fuse-pipe -p fc-agent -p failpoint --check
cargo clippy --all-targets -- -D warnings      # clean

$ make build                                    # ok

     Summary [1259.143s] 116 tests run: 116 passed, 572 skipped

$ make _test-root FILTER="-E 'binary(test_signal_cleanup)'" STREAM=1
     Summary [  33.046s] 9 tests run: 9 passed, 0 skipped
```

All nine signal/cleanup tests pass, including the sudo-hop reaping test (`test_root_test_runner_reaps_vm_when_sudo_is_killed`) that guards the #730 regression.

Ran on a box with no measurement workload active (`pgrep -f "twopath|bench.sh|cdp-rebaseline"` empty), though other agents' builds were running concurrently — these are pass/fail results, not benchmarks.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1013,6 +1013,44 @@ every self-hosted job plus the `scripts/ci-stray-vm-guard.sh` pre/post steps in 
 `always()` cleanup step does not run when the process is SIGKILLed — which is exactly the case
 that leaked. If teardown matters, it must survive SIGKILL.
 
+**pasta is the one hop pdeathsig CANNOT cover — do not "fix" it with a pre_exec.** passt
+changes its own credentials *after* exec: from a bare PID argument, `conf_pasta_ns()` derives
+`/proc/PID/ns/user` and `isolate_user()` does `setns(CLONE_NEWUSER)` into the holder's user
+namespace. That grants capabilities (measured: pasta `CapEff 0x200400` vs fcvm's `0x0`), and
+`commit_creds()` zeroes `task->pdeath_signal` on any credential change that is not a
+capability subset — the same rule that forces `install_namespace_pre_exec`'s hook to run LAST,
+except this transition is inside passt where fcvm cannot re-arm it. Measured with a pre_exec
+pdeathsig installed on pasta: holder gone 2ms after fcvm's SIGKILL, pasta 169-699ms — i.e. the
+hook changed nothing. What actually reaps pasta is passt's own `pasta_netns_quit_*` watchdog:
+`/proc/<holder>/ns/net` disappears when the (pdeathsig'd) holder dies, and passt exits when it
+notices — on procfs it cannot use inotify, so it falls back to a ONE-SECOND interval poll.
+Consequence: pasta is never permanently orphaned, but it can outlive its VM by up to ~1s while
+still holding that VM's forwarded host-side listening sockets, and teardown is now ~40ms, so
+the next VM can be handed the dead VM's loopback IP inside that window. Closing it properly
+means patching passt to re-arm pdeathsig after its isolation completes (`pasta/patches/`, the
+same mechanism as the #661 addr_seen patch). Covered by
+`test_sigkill_kills_firecracker_rootless`, which asserts no PERMANENT orphan.
+
+### TEARDOWN: SIGNAL EVERY CHILD BEFORE WAITING ON ANY
+
+`src/commands/common.rs::cleanup_vm` runs teardown in five phases: stop in-process tasks →
+**SIGKILL the VMM, the holder and the network helper with no await between them** → join their
+exits → network cleanup → synchronous on-disk reaping. The middle rule is load-bearing: the
+kernel reclaims the guest's multi-GiB address space in `exit_mmap()`, charged to the dying
+VMM's system time, and killing-and-waiting each child in turn made every other death queue
+behind it. Never introduce an `.await` between the `start_kill` calls.
+
+**On-disk reaping stays synchronous. Always.** State file, VM data dir, reflink copy, VMM log:
+"leave nothing on disk" is a correctness contract of the same rank as "leave no orphaned VM".
+No janitor, no detached task, no deferral — a process that exits before proving the work is
+done has not done it. It is also not where the time goes: measured `disk_reap_ms` is 54-72ms
+for a full VM's data dir on a loaded host (~10ms for a clone's).
+
+**Report teardown honestly.** `cleanup_vm` logs `caller_blocking_ms`, `processes_gone_ms` and
+`reclaim_cpu_ms` (from `getrusage(RUSAGE_CHILDREN)`) as three separate numbers. Overlapping
+teardown with other work moves the kernel's reclaim cost; it does not delete it. Never quote
+the caller-blocking number alone as if teardown were free.
+
 ### Container Build Rules
 
 **Container builds work naturally with layer caching.** No workarounds needed.
--- a/src/commands/common.rs
+++ b/src/commands/common.rs
@@ -172,13 +172,11 @@ pub async fn spawn_namespace_holder(
 
         // Kill holder if parent (fcvm) dies. Without this, holders orphan to init
         // and accumulate when fcvm is SIGKILL'd or crashes.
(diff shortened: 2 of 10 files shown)


ejc3:
Superseded by #735, which carries the same teardown work plus two things this branch does not: the codex/CodeRabbit fixes (`io::Error::other()` heap-allocating inside the post-fork `pre_exec` — a malloc-after-fork deadlock in exactly the race the check exists to handle, now `from_raw_os_error(ESRCH)`; and the mandatory-discovery early-returns that leaked a live microVM, now `graceful_kill` + `wait`), and the corrected pasta severity.

I opened this before seeing that decision — closing rather than leaving a duplicate open against main, since merging it would regress those two fixes. The branch `fast-teardown` (c1b4fdc2) stays pushed; nothing is lost.

ejc3:
## Retraction: the central claim in this PR's description is wrong

This PR (closed, superseded by #735) asserted in its description, its commit message, and a ~25-line comment in `src/network/pasta.rs` that **"PASTA CANNOT USE PDEATHSIG"** — that a `pre_exec` pdeathsig on pasta is a no-op because passt's `setns(CLONE_NEWUSER)` changes credentials and `commit_creds()` zeroes `task->pdeath_signal`. I backed it with a table showing pasta dying in 169-699 ms with and without the hook, i.e. "unchanged".

**That is false.** Tested properly on #735 ([detail](https://github.com/ejc3/fcvm/pull/735#issuecomment-5225415570)): with the namespace holder held **alive** so passt's PID watch cannot fire, armed pasta dies **2-54 ms** after its parent is SIGKILLed, and unarmed pasta **survives 5 s**. The pdeathsig works.

Two distinct errors, both mine:

1. **The measurement could not attribute cause.** I killed fcvm and the holder together, so passt's 1-second PID watch was live in every run and dominated the result. With both mechanisms armed at once, "unchanged" was the only outcome available — the experiment could not have detected the effect it was testing. There was no negative control.

2. **The mechanism was reasoned from the wrong baseline.** I compared pasta's `CapEff` (`0x200400`) against *fcvm's* (`0x0`) and called that a capability gain. `cred_cap_issubset()` compares a process against **its own prior credentials**, not against its parent's. Under sudo pasta's prior creds are full root, so entering the userns is a capability **loss** — a subset — and the clearing branch is never taken.

Nothing merged from this branch, and #735 ships the correct behaviour, so there is no code impact. Recording it here because the claim is written down in this PR's description and commit message, and anyone reading them would be misled into removing ten lines that are actually load-bearing.

PR #793, opened 2026-08-10
Capture guest and host state when a CDP request fails

ejc3:
Capture guest-side and host-side state from a clone whose CDP request just
failed, before the harness tears it down.

**This fixes nothing.** It collects evidence for a root cause that is still
unsolved, and it deliberately changes no measurement.

## The unsolved question

An 808-clone run of the shared-nothing benchmark (`reqbench.sh run`, UFFD
backend, one clone restored per request, requests strictly sequential, host
loadavg 0.43-0.67) produced 3 failures. What is established:

- Every failure came from a clone whose ARP-triggering readiness ping got no
  reply. Five clones had no reply and three of those failed; of the 803 whose
  ping replied, none failed.
- The guest was alive throughout. Its Chromium kept writing to the serial
  console for the full 100+ seconds.
- Only the two CDP arms failed. `exec` and `noop` reach the guest over vsock and
  never failed, so the break is on the host to guest IP path.
- The shapes were one `TimeoutError` at CDP stage `connect` (120.1 s) and two
  `ConnectionResetError` at stage `enable` (104.7 s, 107.0 s). A healthy request
  finishes in 0.24-0.55 s, so these are hard hangs to the client's own deadline.
- Ruled out with evidence: host load, endpoint collision between concurrent
  clones, and the Firecracker log-volume difference.

What is **not** known is why those guests stopped answering on the IP path and
why they never recovered. A transient startup race resolves in milliseconds.
The leading and unproven hypothesis is that the guest's post-restore network
re-initialisation never ran or never finished: fcvm PUTs a `restore-epoch` per
clone, fc-agent's `watch_restore_epoch` reacts by flushing the ARP cache,
sending a gratuitous ARP and re-registering exec, and a clone that misses that
keeps stale IP networking for life while vsock keeps working.

There is a separate, known proximate bug: `src/network/pasta.rs` only LOGS
`ping_replied=` and returns success on `neighbor_is_resolved` alone, and the
port check that follows connects to pasta's host-side listener, which accepts
regardless of guest state. That is being fixed elsewhere and is not this change.

## What this adds

Vsock exec keeps working during the failure, so the broken guest is
interrogable at exactly the moment it is broken. Until now the harness deleted
it first. `reqbench.py` grows a `FailureProbe` that, on a failed CDP request,
writes `<request-name>.probe.json` into the run's results directory next to that
request's clone log, and references it from the JSONL record.

Guest side, in one bounded `fcvm exec --vm` batch: `ip addr`, `ip -s link`,
`ip route`, `ip neigh`, `/proc/net/arp`, `/proc/net/dev`, `ss -ltnp`,
`/proc/net/tcp`, the Chromium process table with full argv, `dmesg | tail -50`,
any `[fc-agent]` lines in the journal, and the guest clock (restore syncs it
from the host, so a guest still at snapshot time is direct evidence the restore
handler never ran). Then a second, explicitly mutating batch that answers "is it
still broken now": an in-guest CDP connect to `127.0.0.1:9222`, an MMDS read of
the restore-epoch the guest can currently see, and an arping of the gateway.

Host side, for the same clone: the state file, the fcvm process tree, pasta's
pid and process state from its per-VM pid file, the holder's network namespace
through `nsenter -U -n` and independently through the holder's own procfs, a
fresh TCP connect to the published port, and the restore and readiness lines
from that clone's own fcvm log.

One healthy clone per run is captured the same way as the control, because a
dump with nothing to compare against proves little. The failing record names the
control's path so the pair is findable from either end.

## How it stays off the measurement

- The probe runs only after `rec["blocking_ms"]` is stamped, on both the success
  and failure paths. That is the caller-visible latency and the only response
  time these arms publish, and nothing in the probe can move it.
- Only failed CDP requests are probed, plus exactly one healthy control.
- The control is preferentially taken on a warmup rep, which analysis discards
  explicitly. If a run has no warmup rep the control still runs and the record
  it perturbs is stamped `probe_perturbed_timings`, with a line on stderr naming
  it, so its `wall_ms` and teardown are excludable rather than silently mixed in.
- `reqanalyze.py` is untouched. No latency is computed differently.

## Bounding

Every probe command runs in its own session with a hard 20 s timeout enforced by
killing the process GROUP, because `fcvm exec`'s connect ladder spans ~54 s and
the exec holds a command running inside the guest. The whole capture has a 60 s
budget checked before each step; a step that does not fit records why it was
skipped. A command given a shortened timeout is stamped `budget_limited`, so a
`timed_out` caused by the probe's own budget is distinguishable from a wedged
guest. Nothing in the probe can raise into the request: a probe error is
recorded on the record as `probe_error`.

## What the dumps will and will not settle

They will separate the two big branches. If the guest holds a stale or
FAILED gateway neighbour entry, its clock is still at snapshot time, and the
clone log has no `[fc-agent] handling restore` line, the restore handler did not
run and the hypothesis is supported. If the restore narration is complete and
the guest's neighbour table is correct, it is refuted for that clone and the
`ip -s link` and `/proc/net/dev` counters on both sides of the tap say which
direction the packets stop in. The in-guest CDP connect separates a browser
fault from a network fault. The MMDS read says whether the guest could see the
epoch at all.

They will not settle when the failure began: the passive capture is taken ~100 s
after the request gave up, so it describes the steady state, and ordering still
needs the clone's log. They cannot show what Firecracker's MMDS server served at
restore time, only what the guest can read back now. The host-side namespace
view is a single snapshot, so a neighbour entry that expired earlier is
invisible. And the active batch mutates what the passive batch just recorded,
which is why it runs last and is labelled.

## Test evidence

14 new deterministic tests in `bench/chromium/test_reqbench.py`; no VM, no
podman, no built image. Three reds were witnessed by removing the fix under test
from the working code and watching the named test fail, one each for the call
site wiring, the error guard and the process-group kill; all three are quoted in
the commit message.

```
base   817ed25b   Ran 105 tests in 105.137s   FAILED (failures=20, errors=4)
branch             Ran 119 tests in 105.935s   FAILED (failures=20, errors=4)
```

Same box, same built binaries. The failing sets are identical (22 distinct
tests, NEW=[] GONE=[]) and every one is pre-existing on the base, from two
causes both root-caused on this host rather than assumed:

- 20 need `/proc/<pid>/task/<tid>/children`, and
  `grep CONFIG_PROC_CHILDREN /boot/config-$(uname -r)` on this kernel reports
  `# CONFIG_PROC_CHILDREN is not set`, so `children_of()` returns `[]` for every
  process and the whole child-attribution family fails.
- 4 need an `fc-agent` at `reqbench.sh`'s hardcoded
  `target/aarch64-unknown-linux-musl/release/fc-agent`, which cannot exist on an
  x86_64 box. The Makefile already derives `MUSL_TARGET` from the host arch;
  `reqbench.sh` does not. That is a real non-hermeticity, it is not this
  change's, and it is left for its own PR.

Generated probe shell passes `bash -n`, `shellcheck -s sh` and
`shellcheck -s bash` at exit 0, and the passive batch was run against this host
to prove every section command parses and returns a real status (13/13 sections,
471 ms). No Rust changed, so `make lint` was not run.

### DIFF
--- a/bench/chromium/AGENTS.md
+++ b/bench/chromium/AGENTS.md
@@ -452,6 +452,46 @@ Caveat: synthetic dirty-anon pages, not Firecracker's MAP_PRIVATE file-backed
 guest memory; clean file-backed pages may unmap cheaper. Corroboration: the
 observed ~78 ms VM teardown for a ~1 GB VM sits on the ~60 ms/GiB line.
 
+## Failure-time evidence capture: what the probe takes, and what it cannot say
+
+An 808-clone run produced 3 CDP failures. Every one came from a clone whose
+ARP-triggering readiness ping got no reply (5 clones had no reply, 3 of them
+failed; of the 803 whose ping replied, none failed), the guest stayed alive for
+the whole 100+ seconds, and the `exec` and `noop` arms, which reach the guest
+over vsock, never failed. **Why the guest stopped answering on the IP path, and
+why it never recovered, is UNSOLVED.** The leading hypothesis, which nothing has
+yet proved, is that the guest's post-restore network re-initialisation never
+ran or never finished for that clone.
+
+`reqbench.py`'s `FailureProbe` exists because vsock exec keeps working while the
+IP path does not, so the broken guest is interrogable at exactly the moment it
+is broken, and the harness used to delete it. On a failed CDP request, and once
+per run on a healthy one as the control, it writes `<request-name>.probe.json`
+next to that request's clone log and references it from the JSONL record.
+
+- Guest, over `fcvm exec --vm`: addresses, routes, the neighbour table,
+  interface counters, listening sockets, the Chromium process table, `dmesg`,
+  the guest clock, an in-guest CDP connect, an MMDS read of the current
+  restore-epoch, and an arping of the gateway.
+- Host, for the same clone: the state file, the fcvm process tree, pasta's pid
+  and process state, the holder's namespace via `nsenter` and via its procfs, a
+  fresh TCP connect to the published port, and the restore and readiness lines
+  from that clone's own fcvm log.
+
+Read the failure dump against the control; a dump with nothing to compare
+against proves little, which is why the control is captured whether or not the
+run ends up failing. Read `role`, `budget_exhausted` and each command's
+`budget_limited` before reading anything else: a `timed_out` step that was
+`budget_limited` is the probe running out of budget, not a wedged guest.
+
+**What it cannot settle.** The passive capture is taken ~100 s after the request
+gave up, so it describes the STEADY state of the failure, not the moment it
+began; ordering questions still need the clone's log. It cannot see anything
+Firecracker's MMDS server did or did not serve, only what the guest can now read
+back. Its host-side namespace view is one `nsenter` snapshot, so it cannot show
+a neighbour entry that expired earlier. And the active section mutates what the
+passive section just recorded, deliberately and last.
+
 ## Claims currently REFUTED — status per `REVIEW.md`
 
 This list and `REVIEW.md` disagreed for a while: everything here said "do not
--- a/bench/chromium/AGENTS.md
+++ b/bench/chromium/AGENTS.md
@@ -480,9 +480,29 @@ next to that request's clone log and references it from the JSONL record.
 
 Read the failure dump against the control; a dump with nothing to compare
 against proves little, which is why the control is captured whether or not the
-run ends up failing. Read `role`, `budget_exhausted` and each command's
+run ends up failing, and why a control whose dump fails to WRITE does not count
+as taken: the next healthy clone gets another go, bounded at three attempts so a
+systematically broken probe cannot tax every healthy rep in the run. Read
+`role`, `budget_exhausted`, `interrupted_by_signal` and each command's
 `budget_limited` before reading anything else: a `timed_out` step that was
-`budget_limited` is the probe running out of budget, not a wedged guest.
+`budget_limited` is the probe running out of budget, not a wedged guest, and a
+short dump carrying `interrupted_by_signal` is one that stood aside for a
+shutdown.
+
+**It stands aside for INT/TERM, and it is bounded by a process-group kill.** The
+harness's signal handler only records the signal, so nothing in the probe is
+interrupted asynchronously; without a check of its own a capture would spend its
+whole budget with the clone still up and the teardown queued behind it, which is
+long enough for a job runner to escalate to SIGKILL and leave behind exactly the
+clone the probe came to explain. A pending signal therefore skips the capture,
+and one arriving mid-capture stops it at the next step. Each probe command runs
+in its own session, so the timeout kills the whole process GROUP rather than the
+`fcvm exec` wrapper alone, and it names that group by the spawned pid rather
+than by asking the leader for it: a wrapper that exits while its child holds the
+pipe open makes `os.getpgid` raise ESRCH exactly when the kill is needed. The
+kill verifies itself, and `group_kill.survivors` on the command's record is what
+it found, with zombies excluded because a corpse keeps a group present without
+being alive in it.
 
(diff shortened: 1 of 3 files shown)


ejc3:
All four fixed, one commit each, every red witnessed before its fix. Pushed as cb8f2e91..e99949ea.

**P1, reap the killed grandchild.** The old assertion waited for the corpse's procfs entry to vanish, which is a property of whoever inherits the orphan, not of the kill. Reproduced your environment with a pid namespace whose init is the test runner (`unshare --pid --fork --mount-proc -r`): `AssertionError: ('Z', 0, 0, 254002254) is not None`, after waiting out the full 5s deadline on a process that had been dead the whole time. The test now becomes a child subreaper, `waitpid`s the orphan and asserts the wait status is SIGKILL, falling back to procfs absence only when PID 1 got there first. That is stronger than the absence it replaced because it also names the signal, and it runs in 1.0s instead of 6.0s. Proved still non-vacuous by swapping the group kill for `proc.kill()`: `pid 4146913 is still running 5s later, state ('S', ...)`. State `S`, not `Z`, is exactly the distinction the old assertion could not make.

**P1, abort capture on a pending signal.** The handler only records the signal and unwinding happens at polls inside the request scope, so the probe sat in the one window with no poll: after the request's last one, before teardown. Reds with each check removed: `2.0055270083248615 not less than 0.5 : the probe ran with a termination signal pending`, and `'active_mutating' unexpectedly found in {...} : the steps after the signal must not run`. Both inject through `record_harness_interrupt`, the harness's own handler, so they run against the state a real INT/TERM leaves; the mid-capture one fires from inside the first probe command, where a capture actually spends its time. `observe` now skips outright, `capture` stops at the next step and stamps `interrupted_by_signal`, and `nsenter_batch` checks between variants. The partial dump is still written, since by then it has cost nothing further.

**P2, retry the healthy control.** `control_captured` was set in a block that ran on the exception path too, so one transient error retired the run's only control and every later failure dump named a control that does not exist. Reds: `True is not false : a dump that was never written is not the control`, and `Lists differ: [0] != [0, 1, 2] : the control retry is not bounded`. The flag now moves to where the dump is written, capped at three attempts because each costs up to the full budget and perturbs the rep it lands on. The perturbation warning names the attempt, and it now prints even when the capture raised, since an attempt perturbs the timings either way.

**P2, kill the stable process group id.** Confirmed the mechanics before fixing: `getpgid` on a *zombie* leader still succeeds, so the ESRCH needs the leader to be **reaped**, which means the trigger is a SIGCHLD reaper. Opened that window deterministically with a wrapper that exits while its child holds the stdout pipe, so `communicate()` still blocks the full timeout. Red: `the group kill missed the descendant the vanished leader left behind: pid 4193142 is still running 5s later, state ('S', ...)`. The fix signals `killpg(proc.pid)`, valid because `start_new_session=True` makes the group id equal the leader pid and a live member pins that number against reuse, then verifies with `live_group_members`. `killpg(pgid, 0)` would not do for the verification: an unreaped zombie keeps the group present, so on a non-reaping PID 1 it would report a dead group alive forever. Second red for that helper: `70681 unexpectedly found in [70680, 70681, 70682] : a zombie is dead, so it is not a survivor`.

**Gate.** `make test-chromium-request` on one box, same binaries:

```
origin/main   72ace3fd   105 tests   FAILED (failures=20, errors=4)
branch before            119 tests   FAILED (failures=20, errors=4)
branch after             125 tests   FAILED (failures=20, errors=4)
```

NEW=[] GONE=[] against both. The 22 distinct failures were bucketed rather than assumed: 20 are child attribution (this kernel reports `# CONFIG_PROC_CHILDREN is not set`), 4 are the hardcoded aarch64-musl `fc-agent` path, 0 other. The probe suites are also green under a non-reaping PID 1 (24 tests, OK). Shell wrapper and generated batches pass `bash -n` and `shellcheck` under both `-s sh` and `-s bash`. No Rust changed.

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

PR #794, opened 2026-08-10
Fix five review findings on the merged benchmark changes

ejc3:
`make bench-quick` shipped in #790 unable to run. Its recipe expanded criterion's
flags straight onto cargo's own command line, and cargo rejects the first one
before any benchmark starts:

```
$ make bench-quick
... cargo bench -p fuse-pipe --bench throughput --sample-size 10 --warm-up-time 1 --measurement-time 3
error: unexpected argument '--sample-size' found
  tip: to pass '--sample-size' as a value, use '-- --sample-size'
make: *** [Makefile:691: bench-throughput] Error 1
```

That is the same defect #790 set out to remove, a documented target that does not
work, and it got there because the recipe was checked with `make -n`. A dry run
prints a command line; it does not ask cargo whether the command line is one it
would accept. This PR is five review findings on that PR, all read after it
merged, and every fix here is verified by running the thing rather than printing
it.

**Stacked on:** main.

### `make bench-quick` reaches cargo correctly

Criterion's flags now follow `--`. The separator is omitted when there is nothing
behind it, so plain `make bench` is unchanged.

```
$ make bench-quick
==> Running fuse-pipe benchmarks: throughput, operations, protocol
     Running benches/throughput.rs (target/release/deps/throughput-5ba9d7c400bdd285)
     Running benches/operations.rs (target/release/deps/operations-8df94d73b357a580)
     Running benches/protocol.rs  (target/release/deps/protocol-b3bff7af4daa0dc5)
MAKE EXIT=0
```

Note while you are in here: a group that calls `sample_size()` in code keeps its
own count, so `--sample-size` reaches only the groups that do not.
`--warm-up-time` and `--measurement-time` reach all of them. The recipe comment
says so.

### `make -j bench` no longer runs the three suites at once

They were prerequisites, and make parallelizes prerequisites. `bench` is now one
target with three recipe lines, which make never parallelizes, so neither an
explicit `-j` nor one inherited through `MAKEFLAGS` can reach it. The suites stay
separate targets sharing one canned recipe, so nothing is duplicated.

GNU Make 4.3 is what this repo builds on and neither of make's own mechanisms
works there: `.NOTPARALLEL` ignores its prerequisites before 4.4 and serializes
the whole makefile rather than one target, and `.WAIT` arrived in 4.4.

### The ownership repair can now both fire and fail

Two problems, and the second is why the first never mattered.

It kept cargo's exit status across the `chown`, so a repair that could not run
reported the suite green.

It also scanned a directory criterion never writes to. Criterion's second choice
of output directory is `$CARGO_TARGET_DIR/criterion`, and `CARGO_TARGET_DIR` here
is the relative string `target`, which criterion resolves inside the bench
binary, whose working directory cargo sets to the package root:

```
$ make bench-quick                                    # exit 0
$ find target/ -user root -print -quit                # the repair scans here: nothing
$ find fuse-pipe/target/criterion -user root -maxdepth 1
fuse-pipe/target/criterion/single_op_getattr
fuse-pipe/target/criterion/parallel_reads
fuse-pipe/target/criterion/parallel_writes

$ sudo chown -R root:root fuse-pipe/target/criterion  # what a first privileged run leaves
$ make bench-protocol                                 # exit 0
Criterion.rs ERROR: error: Failed to access file
  "target/criterion/serialize_lookup_request/new/sample.json": Permission denied (os error 13)
```

`CRITERION_HOME` now pins the output to one absolute path, the repair scans and
repairs that path, and both the scan and the `chown` decide the recipe's exit
status. After the fix:

```
$ make bench-protocol
0 permission errors
$ ls target/criterion/serialize_lookup_request/new/
benchmark.json  estimates.json  sample.json  tukey.json
$ make bench-operations && find target/criterion -user root -print -quit
                                # empty: the repair fired and succeeded
```

### `single_op/getattr` measures GETATTR

The `_uncached` case did not. For a path the mount has not seen, pathname
resolution sends a LOOKUP whose reply already carries attributes, so the stat is
satisfied and no GETATTR is issued. It measured cold path resolution, which is
what the lookup group is for.

Isolating GETATTR needs a descriptor, so there is no name to resolve, and a
forced revalidation: `statx(fd, "", AT_EMPTY_PATH|AT_STATX_FORCE_SYNC, ...)`
makes fuse call `fuse_do_getattr()` instead of answering from the attribute
cache. The case is now `fuse_256_readers_forced_round_trip`, with
`host_fs_forced` beside it to price the syscall and the flag where there is
nothing to revalidate.

### Neither round-trip case can silently revert to measuring a cache

The pool of 20,000 distinct files is gone. `sample_size` bounds criterion's
samples, not its `iter` calls, so sizing a pool against it bounds nothing.
Measured on the merged version at criterion's defaults, the case ran 20,000
measurement iterations against that pool, after an untimed warm-up that had
already walked thousands of entries. Past the wrap, whether a revisited inode is
still cached depends on how long one pass takes against the 1s `entry_timeout`,
which is a property of the host rather than of the benchmark, and the case cannot
report which blend it got:

```
single_op/getattr/fuse_256_readers_attr_cache_hit   17M iterations   [297.38 ns .. 310.12 ns]
single_op/getattr/fuse_256_readers_uncached         20k iterations   [181.50 µs .. 465.15 µs]
```

The replacements have no such dependence. A forced statx cannot be served from
the attribute cache, and a name that does not exist is never cached, because
fuse-pipe answers a miss with a plain ENOENT and `fuse_lookup()` invalidates the
entry rather than caching it.

Both cases also assert they reach the server before measuring anything. Chmod the
backing file behind the mount's back and the forced statx must report the new
mode while a plain stat still reports the old one; create the probe name behind
the mount's back and the mount must see it appear. The first of these earned its
place on its first run, failing because the probe read the file's size:
fuse-pipe negotiates `FUSE_WRITEBACK_CACHE`, under which the kernel owns `i_size`
for regular files and discards the server's. Permission bits come out of every
reply.

```
single_op/getattr/host_fs                              283.80 ns
single_op/getattr/host_fs_forced                       153.72 ns
single_op/getattr/fuse_256_readers_attr_cache_hit      297.90 ns
single_op/getattr/fuse_256_readers_forced_round_trip    83.143 µs
single_op/lookup/host_fs                               303.47 ns
single_op/lookup/host_fs_miss                          184.50 ns
single_op/lookup/fuse_256_readers_dentry_cache_hit     291.60 ns
single_op/lookup/fuse_256_readers_miss_round_trip      120.33 µs
```

x86_64 devserver, 3s windows, not the c7gd.metal the published table came from,
so these are shapes rather than a replacement for it. The cache-hit cases sit on
the host baseline, which is where the published 1.05x came from; the round trips
cost 293x and 396x it. PERFORMANCE.md's two rows are relabelled for the caches
they measure, and its "metadata ops have ~5% overhead" conclusion is replaced by
what those rows are plus a pointer to the round-trip cases.

## Tests

`tests/test_documented_make_targets.rs` gains a harness that runs the real
Makefile with a stub in place of cargo, plus find and sudo stubs that set up the
ownership-repair precondition without needing a root-owned file and keep a unit
test from chowning a developer's tree. It skips `cargo-target-link`, which takes
the target generation's lease exclusively while the `cargo nextest` running the
test holds that lease shared; a harness that calls it deadlocks against its own
test runner, and did, for 120s, five times, before that was fixed.

Every new guard was watched failing on the merged code first.

Against the merged Makefile:

```
$ git checkout origin/main -- Makefile
$ cargo nextest run --release --no-default-features --test test_documented_make_targets
FAIL bench_quick_passes_criterion_flags_after_the_cargo_separator
  bench-quick puts criterion flags on cargo's own command line, where cargo
  rejects them with `unexpected argument`:
  bench -p fuse-pipe --bench throughput --sample-size 10 --warm-up-time 1 --measurement-time 3
FAIL parallel_make_never_runs_two_benchmark_suites_at_once
  `make -j8 bench` ran benchmark suites concurrently:
  ["bench -p fuse-pipe --bench protocol", "bench -p fuse-pipe --bench operations"]
FAIL a_failed_ownership_repair_fails_the_privileged_bench_target
  bench-throughput reported success while the ownership repair failed
FAIL bench_recipes_pin_criterion_output_to_an_absolute_path
  left: "<unset>"  right: "<repo>/target/criterion"
Summary: 8 tests run: 4 passed, 4 failed
```

Against the merged `operations.rs`:

```
$ git checkout origin/main -- fuse-pipe/benches/operations.rs
FAIL metadata_benchmarks_name_the_cache_they_measure
  bench_getattr has no `_round_trip` case, so nothing in it measures a FUSE
  round trip: a repeated path is answered by the kernel's attribute/dentry cache
FAIL metadata_benchmarks_cannot_wrap_a_fixed_pool
  operations.rs is back on a fixed path pool (METADATA_POOL)
Summary: 8 tests run: 6 passed, 2 failed
```

Green with the fixes:

```
$ cargo nextest run --release --no-default-features --test test_documented_make_targets
Summary [3.4s] 8 tests run: 8 passed, 0 skipped
```

## Gates

```
$ cargo fmt --all -- --check                      # clean
$ cargo clippy --all-targets -- -D warnings       # clean
$ make test-unit
Summary [51.0s] 622 tests run: 615 passed, 7 failed, 0 skipped
```

The 7 are `uffd::` tests, and they fail the same way on untouched `origin/main`
(`68 tests run: 61 passed, 7 failed`, same names) on this host, which has
`vm.unprivileged_userfaultfd=0`:

```
creating userfaultfd (via /dev/userfaultfd):
  OpenDevUserfaultfd(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })
```

`bench-throughput` and `bench-operations` mount a real FUSE filesystem under the
privileged runner and both ran here, so every claim above is from a real run.

## Two pre-existing things noticed while running this, neither touched here

`src/uffd/server.rs`'s `spawn_victim()` starts `sleep 600` as a stand-in VMM. On
a host where the uffd tests fail before killing it, that child is orphaned and
survives, and it inherits the shared lease that `cargo-target-run.sh` holds on
the worktree's cargo target generation. `cargo-target-link.sh` then blocks on its
exclusive lease, so the next `make` sits in `build` until the sleep expires:

```
2011850 fd=10 -> /mnt/fcvm-btrfs/cargo-target/wt-bfix-...  scripts/cargo-target-link.sh
2011887 fd=10 -> /mnt/fcvm-btrfs/cargo-target/wt-bfix-...  flock -x 10
1971440 fd=11 -> /mnt/fcvm-btrfs/cargo-target/wt-bfix-...  sleep 600   (orphan, PPID 1)
1975387 fd=11 -> /mnt/fcvm-btrfs/cargo-target/wt-bfix-...  sleep 600   (orphan, PPID 1)
```

A five-minute wedge cleared the moment those two were killed. It needs the uffd
tests to be failing to happen at all, so it does not reach CI, but a victim that
outlives its test should not be holding a build lease.

`tests/test_serve.rs` warns `unused import: std::time::Duration` under
`--no-default-features`, because its uses sit behind
`#[cfg(feature = "integration-slow")]`.

## Summary by CodeRabbit

- **New Features**
  - Improved benchmark command handling, including clearer separation of benchmark options and a configurable results location.
  - Benchmark suites now run sequentially and safely handle privileged operations.
  - Added validation that benchmark results are saved successfully and remain accessible.

- **Bug Fixes**
  - Corrected metadata benchmarks to measure actual server round trips instead of cache hits.
  - Improved handling of benchmark result ownership and failure reporting.

- **Documentation**
  - Updated performance documentation to distinguish cached operations from server-reaching round trips and reflect the revised measurements.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -717,15 +717,20 @@ define run_privileged_bench
 	CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER='$(ROOT_TEST_RUNNER)' \
 	$(CARGO) bench -p fuse-pipe --bench $(1) $(BENCH_SEPARATED); \
 	bench_rc=$$?; \
-	if [ -d '$(CRITERION_HOME)' ]; then \
-		if ! root_owned=$$(find '$(CRITERION_HOME)' -user root -print -quit); then \
-			echo "ERROR: cannot scan $(CRITERION_HOME) for root-owned files, so the ownership repair after --bench $(1) never ran" >&2; \
+	if [ ! -d '$(CRITERION_HOME)' ]; then \
+		if [ $$bench_rc -eq 0 ]; then \
+			echo "ERROR: --bench $(1) exited 0 but $(CRITERION_HOME) does not exist. criterion logs persistence failures and still returns 0, so the suite printed timings and saved no sample.json, estimates.json or baseline: nothing can be compared and no regression can ever be reported" >&2; \
 			exit 1; \
 		fi; \
-		if [ -n "$$root_owned" ] && ! sudo chown -R $$(id -u):$$(id -g) '$(CRITERION_HOME)'; then \
-			echo "ERROR: $(CRITERION_HOME) is still root-owned after --bench $(1); bench-protocol cannot persist criterion output and _test-root will refuse to start" >&2; \
-			exit 1; \
-		fi; \
+		exit $$bench_rc; \
+	fi; \
+	if ! root_owned=$$(find '$(CRITERION_HOME)' -user root -print -quit); then \
+		echo "ERROR: cannot scan $(CRITERION_HOME) for root-owned files, so the ownership repair after --bench $(1) never ran" >&2; \
+		exit 1; \
+	fi; \
+	if [ -n "$$root_owned" ] && ! sudo chown -R $$(id -u):$$(id -g) '$(CRITERION_HOME)'; then \
+		echo "ERROR: $(CRITERION_HOME) is still root-owned after --bench $(1); bench-protocol cannot persist criterion output and _test-root will refuse to start" >&2; \
+		exit 1; \
 	fi; \
 	exit $$bench_rc
 endef
@@ -733,7 +738,13 @@ endef
 # One criterion suite as the invoking user. Nothing here writes as root, so
 # there is no ownership to repair.
 define run_unprivileged_bench
-	CRITERION_HOME='$(CRITERION_HOME)' $(CARGO) bench -p fuse-pipe --bench $(1) $(BENCH_SEPARATED)
+	CRITERION_HOME='$(CRITERION_HOME)' $(CARGO) bench -p fuse-pipe --bench $(1) $(BENCH_SEPARATED); \
+	bench_rc=$$?; \
+	if [ ! -d '$(CRITERION_HOME)' ] && [ $$bench_rc -eq 0 ]; then \
+		echo "ERROR: --bench $(1) exited 0 but $(CRITERION_HOME) does not exist. criterion logs persistence failures and still returns 0, so the suite printed timings and saved no sample.json, estimates.json or baseline: nothing can be compared and no regression can ever be reported" >&2; \
+		exit 1; \
+	fi; \
+	exit $$bench_rc
 endef
 
 # The three fuse-pipe suites, in order.
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,31 @@ export CARGO_TARGET_DIR := target
 NEXTEST := $(CARGO) nextest $(NEXTEST_CMD) --release
 TEST_CONFIG_WRAPPER := ./scripts/with-test-config.sh
 # Extra flags forwarded to every criterion bench recipe (see bench-quick).
+#
+# Criterion's flags belong to the BENCH BINARY, so they have to follow `--`.
+# Given to cargo directly they are rejected by its own argument parser before
+# any benchmark starts:
+#
+#     $ cargo bench -p fuse-pipe --bench throughput --sample-size 10
+#     error: unexpected argument '--sample-size' found
+#       tip: to pass '--sample-size' as a value, use '-- --sample-size'
+#     Usage: cargo bench --package [<SPEC>] --bench [<NAME>] [-- [ARGS]...]
+#
+# BENCH_SEPARATED is what the recipes interpolate, and it stays empty when
+# BENCH_ARGS is, so plain `make bench` carries no trailing separator.
 BENCH_ARGS ?=
+BENCH_SEPARATED = $(if $(strip $(BENCH_ARGS)),-- $(BENCH_ARGS))
+
+# Where criterion keeps baselines and reports. Pinned to an absolute path
+# because criterion's default is not the directory it looks like.
+#
+# Its second choice after CRITERION_HOME is `$CARGO_TARGET_DIR/criterion`, and
+# CARGO_TARGET_DIR here is the relative string `target`, which criterion
+# resolves inside the bench binary, whose working directory cargo sets to the
+# package root. Every suite was therefore writing to `fuse-pipe/target/criterion`
+# while the ownership repair scanned `target/` at the repo root and found
+# nothing to do. Overridable so tests can point it at a scratch directory.
+CRITERION_HOME ?= $(CURDIR)/target/criterion
 
 # cargo/nextest target runner for the privileged suites. cargo-nextest stays unprivileged
 # and only the test binary is elevated, so the `sudo` hop sits in the middle of the process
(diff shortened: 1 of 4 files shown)


ejc3:
Fixed in 2a627a6f: both bench macros now fail when the criterion output directory is absent after a zero-exit run, with a witnessed red on the unfixed Makefile and a control that passes when the directory exists. The `[ -d ... ]` wrapper around the ownership block is gone, since by that point the directory is known to exist.

ejc3:
Both findings folded in 4a80570e: the cache-liveness half of `assert_forced_getattr_reaches_server` now retries up to five times so an expired attr_timeout reads as a retry instead of a benchmark abort (the forced half stays hard), and the persistence message literal is joined with line continuations. `make test-unit FILTER=bench`: 7 passed.

PR #795, opened 2026-08-10
Salvage the unlanded work from the archived WIP branches

ejc3:
# Salvage six items out of the archived WIP branches

Each commit stands alone and can be dropped without disturbing the others. Six
archived branches were reduced to what is actually worth keeping; the branches
themselves carry nothing else.

## 1. A live bug on main: the health monitor was detached, not cancelled

`cleanup_vm` moved the health-monitor `JoinHandle` into a `tokio::select!` racing
a 100 ms sleep. Dropping a tokio `JoinHandle` **detaches** its task rather than
cancelling it, so whenever the sleep arm won, a health monitor that had not yet
observed its cancellation token kept running while cleanup proceeded to destroy
its network namespace.

This is the same defect as the one already fixed in `wait_for_stderr_eof`
(`src/utils.rs`), which uses the borrow-then-abort shape. `stop_health_monitor`
now does the same: `tokio::time::timeout(budget, &mut handle)` keeps ownership,
so the budget can expire without consuming the handle, and the task is aborted
**and awaited** before teardown continues.

Witnessed red against the unfixed code:

```
$ make _test-unit FILTER="-E 'test(=commands::common::tests::health_monitor_timeout_aborts_and_reaps_task)'"
thread '...health_monitor_timeout_aborts_and_reaps_task' panicked at src/commands/common.rs:3343:9:
the timed-out health-monitor future must be dropped before the stop barrier returns
Summary [   5.015s] 1 test run: 0 passed, 1 failed, 616 skipped

# after the fix
Summary [   0.009s] 1 test run: 1 passed, 616 skipped
```

### Other instances of the same shape

The whole repo was swept for a `JoinHandle` consumed by value in a construct
whose other arm can win. Nine sites match the shape; none is the same bug, and
the reasoning for each is in the review notes below.

## 2. Two tests that close a real hole in `wait_for_stderr_eof`

The fix landed inline on main but its tests did not, and main covers only the
timeout branch. An implementation that abandoned the reader immediately would
satisfy the existing test while silently truncating the stderr tail on every
error path that renders one.

Demonstrated by stubbing the function to abort and return without waiting: the
**existing** test still passes, and the new one fails.

```
# stubbed
PASS [   0.206s] (2/3) fcvm utils::tests::stderr_eof_timeout_aborts_the_reader
TRY 2 FAIL [   0.007s] (3/3) fcvm utils::tests::wait_for_stderr_eof_waits_for_delayed_bulk_output
  stderr waiter returned while the writer still held the pipe open
Summary [   5.016s] 3 tests run: 2 passed, 1 failed, 616 skipped

# stub removed
Summary [   0.261s] 3 tests run: 3 passed, 616 skipped
```

## 3. The instrument behind the headline UFFD numbers

`src/uffd/working_set.rs` opens by citing a cross-clone Jaccard median of 0.927,
~56,300 faults per Chromium clone, and 8.6% sequential faults. Nothing in the
tree could re-derive any of it. `FaultTrace` plus `faultbench.py` and
`faultanalyze.py` are what produced those figures.

Hand-ported rather than cherry-picked: `server.rs` was restructured around
`PeerVmm`/`VmContext`/`VmState` since. The trace lives in `VmState`, reuses its
existing `started` instant, and hooks the four points in `drain_events` where a
fault is actually resolved.

The measurement needs a golden snapshot and a working userfaultfd, so the
runnable claim is the wire format between the Rust writer and the Python reader
(`struct.unpack_from("<QQQ", data, i * 24)`) — pinned by a test, witnessed red by
flipping the writer to `to_be_bytes`.

## 4. An open-loop scale harness

`bench-chromium-request` cannot drive an arrival rate: its driver is
`for rep in range(...)`, so offered load is whatever the system served and a
slower backend is handed a lighter load. It also compares FILE and UFFD across
separate runs, confounding the backend with drift — the defect
`bench/chromium/AGENTS.md` records as "interleave, never block".

`reqscale.py` launches on absolute deadlines and interleaves both backends
*within* each rate interval. Four make targets, written to this Makefile's
conventions rather than the ones the harness shipped with, so the existing
`test-chromium-request` is left intact and `FILTER=` keeps working.

`test_reqscale.py` arrived with five fixture bugs, each a test that could never
reach the assertion it advertised (an `assertIn` against an argv list, a rate
that tripped an earlier guard, a terminal sample whose nested timestamps stayed
behind, a physically impossible `wall_ms`, and an expectation of 1 failure from a
spec that plans a pair). All five fail identically on the branch they came from,
so none is port damage.

```
$ make test-chromium-scale
Ran 65 tests in 31.110s
OK
```

Every new target was run, not dry-run — a target has shipped broken before
because it was only checked with `make -n`.

## 5. One runner and two documentation gaps

`test-clone-floor-overlap` runs three concurrent nextest lanes. A single nextest
process cannot produce the interleaving, because `stress-tests` and
`hugepage-tests` are pinned to `max-threads = 1`.

Also documents `FCVM_UFFD_MAX_CLONES` (default **256**; it was the only `FCVM_*`
variable living solely in `src/`) and records why UFFD queue depth is not
observable today, so `agent/archive-uffd-profile-20260810` can be deleted.

## Gates

```
$ make lint          # fmt, clippy -D warnings, audit, deny — all clean, exit 0

$ make test-unit
Summary [  50.744s] 621 tests run: 614 passed, 7 failed, 0 skipped
```

The four added tests all pass, and the run that selects only them reports
`617 skipped`, which is the base's own test count:

```
PASS fcvm utils::tests::wait_for_stderr_eof_accepts_a_late_waiter_after_eof
PASS fcvm utils::tests::wait_for_stderr_eof_waits_for_delayed_bulk_output
PASS fcvm commands::common::tests::health_monitor_timeout_aborts_and_reaps_task
PASS fcvm uffd::server::tests::fault_trace_writes_the_little_endian_triples_faultanalyze_reads
Summary [   0.009s] 4 tests run: 4 passed, 617 skipped
```

617 + 4 = 621, so the series adds exactly four tests and four passes. The
failing set is **byte-identical** to the baseline measured on untouched main
(`616 tests run: 609 passed, 7 failed` at the older tip) — seven uffd tests that
need `vm.unprivileged_userfaultfd`, which is `0` on this host.

```
$ make test-chromium-request
Ran 105 tests in 104.802s
FAILED (failures=20, errors=4)
```

Also identical to this host's baseline, measured with the touched files reverted
(`Ran 105 tests in 103.876s / FAILED (failures=20, errors=4)`, same test names).
Every one is a process-forking test failing with "pid N never forked a child".

## Summary by CodeRabbit

- **New Features**
  - Added Chromium request-scalability benchmarking with JSON analysis, Markdown reports, and reproducible tests.
  - Added fault benchmarking and reporting for memory usage, latency, locality, stability, and fault activity.
  - Added optional UFFD fault tracing through `FCVM_UFFD_FAULT_TRACE`.
  - Added `FCVM_UFFD_MAX_CLONES`, defaulting to 256 clones per snapshot-serving process.

- **Bug Fixes**
  - Improved Chromium page cleanup and readiness validation.
  - Improved health-monitor shutdown and delayed output handling.
  - Added stronger process and cgroup cleanup safeguards.

- **Documentation**
  - Expanded CLI and benchmarking documentation, including tracing limitations and health checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,6 +231,7 @@ CONTAINER_RUN := $(CONTAINER_RUN_BASE) --ulimit nproc=65536:65536 --pids-limit=6
 	cargo-target-link build-host-tools setup-btrfs setup-default setup-fcvm setup-pjdfstest setup-hugepages bench bench-vm bench-hugepages bench-hugepages-test \
 	bench-container-import bench-chromium bench-chromium-request analyze-chromium-request bench-clone-latency test-chromium-request \
 	bench-chromium-scale analyze-chromium-scale report-chromium-scale test-chromium-scale \
+	test-chromium-fault \
 	bench-quick bench-throughput bench-operations bench-protocol \
 	lint fmt update-dependency ssh test-serve-sdk
 
@@ -295,6 +296,7 @@ help:
 	@echo "  analyze-chromium-scale  Validate a scale run and write deterministic JSON"
 	@echo "  report-chromium-scale   Validate a scale run and write plain Markdown"
 	@echo "  test-chromium-scale  Run the scale harness's deterministic unit tests"
+	@echo "  test-chromium-fault  Run the fault harness's deterministic unit tests"
 	@echo "  bench-clone-latency  Clone spawn->exec-ready latency (LABEL=, N=)"
 	@echo ""
 	@echo "CI merge train (pooled CI for a batch of PRs, see docs/ci-train.md):"
@@ -697,6 +699,10 @@ test-chromium-request:
 
 test-chromium-scale:
 	@python3 -m unittest discover -s bench/chromium -p 'test_reqscale.py' \
+		$(if $(FILTER),-k '$(FILTER)',)
+
+test-chromium-fault:
+	@python3 -m unittest discover -s bench/chromium -p 'test_faultbench.py' \
 		$(if $(FILTER),-k '$(FILTER)',)
 
 # Open-loop CDP-fast scalability and page-fault measurement. Unlike
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ CONTAINER_RUN_BASE := podman run --rm --privileged \
 CONTAINER_RUN := $(CONTAINER_RUN_BASE) --ulimit nproc=65536:65536 --pids-limit=65536
 
 .PHONY: all help build clean clean-test-data check-disk \
-	test test-unit test-fast test-all test-root test-packaging test-ci-infrastructure fuzz \
+	test test-unit test-fast test-all test-root test-packaging test-ci-infrastructure test-clone-floor-overlap fuzz \
 	_test-unit _test-fast _test-all _test-root _setup-fcvm _bench \
 	container-build container-test container-test-unit container-test-fast container-test-all container-test-fc-mock \
 	container-setup-fcvm container-shell container-clean container-bench \
@@ -251,6 +251,7 @@ help:
 	@echo "  test-all           + slow VM tests (rootless, no sudo)"
 	@echo "  test-root, test    + privileged tests (bridged, pjdfstest, sudo)"
 	@echo "  test-fc-mock       Run tests with fc-mock (no KVM required)"
+	@echo "  test-clone-floor-overlap  Reproduce the clone/CH/hugepage lifecycle overlap"
 	@echo "  test-ci-infrastructure  Deterministic runner-failure classifier fixtures"
 	@echo "  fuzz               Seeded lifecycle chaos fuzz (SEEDS=N|list OPS=M, defaults 1/10)"
 	@echo ""
@@ -744,6 +745,32 @@ report-chromium-scale:
 	@test -n "$(SCALE_REPORT)" || (echo "ERROR: SCALE_REPORT required"; exit 1)
 	python3 bench/chromium/reqscale_analyze.py \
 		--run-dir "$(SCALE_RUN_DIR)" --markdown-out "$(SCALE_REPORT)"
+
+# Run three ordinary nextest lanes concurrently. Each lane keeps the repository's
+# normal scheduling (in particular, hugepage tests remain serialized by their
+# configured test group); separate lanes are needed on one-CPU test hosts where a
+# single nextest process cannot overlap the clone stress test with lifecycle tests.
+# Standard helpers keep writing their per-fcvm debug logs, while each lane also
+# captures its complete nextest and cleanup transcript.
+CLONE_FLOOR_CLONE_FILTER := -E 'package(fcvm) & test(=test_snapshot_clone_stress_100_bridged)'
+CLONE_FLOOR_CH_FILTER := -E 'package(fcvm) & test(=test_cloud_hypervisor_cold_boot)'
+CLONE_FLOOR_HUGEPAGE_FILTER := -E 'package(fcvm) & (test(=test_hugepage_vm_boot) | test(=test_hugepage_cache_restore_uses_uffd) | test(=test_hugepage_snapshot_clone))'
+test-clone-floor-overlap: show-notes check-disk setup-fcvm setup-hugepages
+	@mkdir -p $(TEST_LOG_DIR)
+	@set -uo pipefail; \
+		run_lane() { \
+			lane="$$1"; filter="$$2"; \
+			$(MAKE) --no-print-directory _test-root STREAM=1 FILTER="$$filter" 2>&1 | \
+				tee "$(TEST_LOG_DIR)/clone-floor-overlap-$${lane}.log"; \
+		}; \
+		run_lane clones "$(CLONE_FLOOR_CLONE_FILTER)" & clones_job=$$!; \
+		run_lane cloud-hypervisor "$(CLONE_FLOOR_CH_FILTER)" & ch_job=$$!; \
+		run_lane hugepages "$(CLONE_FLOOR_HUGEPAGE_FILTER)" & hugepages_job=$$!; \
+		clones_rc=0; wait "$$clones_job" || clones_rc=$$?; \
+		ch_rc=0; wait "$$ch_job" || ch_rc=$$?; \
+		hugepages_rc=0; wait "$$hugepages_job" || hugepages_rc=$$?; \
+		echo "clone-floor overlap results: clones=$$clones_rc cloud-hypervisor=$$ch_rc hugepages=$$hugepages_rc"; \
+		test "$$clones_rc" -eq 0 -a "$$ch_rc" -eq 0 -a "$$hugepages_rc" -eq 0
 
 test-ci-infrastructure:
 	@PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s tests -p 'test_ci_infrastructure.py'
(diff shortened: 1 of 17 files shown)


ejc3:
All five fixed in ae83fb39, each with a witnessed red. They are one defect class: an instrument that attributes a cost to a specific request, reporting something plausible where it should report nothing.

**P1, ambiguous fault-trace matches.** Confirmed there is no identity to resolve the tie with: the trace filename is `<serve_pid>-<vm_id>.faults`, and that `vm_id` is the serve process's own per-connection counter (`src/uffd/server.rs`, `let vm_id = format!("vm-{}", next_vm_id)`), not the clone name the harness knows. Binding exactly would need a protocol change, so this takes your first option: `match_trace` returns None on two or more candidates and stamps `trace_ambiguous`. `agg` already guards with `.get`, so the result is a smaller `n` rather than a corrupt mean. Red: `PosixPath('.../1000-vm-1.faults') is not None : an ambiguous window must not be attributed to either`.

**P1, discarded `wait_clones_gone`.** `require_clones_gone` raises instead. Stopping rather than skipping, because the records already written are valid and only the ones after the surviving clone are not. Red: `SystemExit not raised`.

**P2, `--out` reuse.** `require_fresh_out_dir` refuses a non-empty directory; absent or empty is fine. Red: `SystemExit not raised`.

**P2, untimed CONTINUE retry.** The parked fault now carries its trace start, and the retry that actually releases the vCPU closes the interval. Red: `parking must not close the interval; the vCPU is still blocked`.

**CodeRabbit's `ZeroDivisionError`.** Fixed, with `None` rather than `0` for the undefined case: a fraction of an empty mean set is not zero. Red: `ZeroDivisionError: float division by zero`.

**Testability.** The fault harness had no tests and no make target, so the four Python rules were extracted into `match_trace`, `require_clones_gone` and `require_fresh_out_dir`, and the two Rust bookkeeping steps into `park_continue` and `close_parked_trace`. That makes all five testable with no VM and no `/dev/userfaultfd`. New: `bench/chromium/test_faultbench.py` (10 tests) and `make test-chromium-fault`. Each red above was produced by restoring one pre-fix behaviour, and the six positive controls stayed green throughout, including the one that pins the grace period (a trace landing 3s after `t1` is still this run's).

`bench/chromium/AGENTS.md` gained a section stating the four rules, since the harness had no documentation at all.

Gate: `cargo fmt`, `cargo clippy --all-targets` clean. `test_reqbench` is 105 tests / 20 failures / 4 errors, which is the identical set `origin/main` produces on this host: 20 need `CONFIG_PROC_CHILDREN` (not set in this kernel) and 4 hardcode an aarch64-musl `fc-agent` path.

ejc3:
The remaining 13 findings are fixed in bf51fedc. (My first sweep queried `reviewThreads(first:30)` before CodeRabbit had finished posting and saw only 5, which is the truncation trap: 18 threads exist.)

**Attribution.** Filtering kvm events on `r["fc_pid"]` as proposed would have discarded *everything*: the tracepoint fires in vCPU thread context, so its pid field is a TID and never equals the process id. The sampler now records the firecracker's thread ids while the request runs (`/proc/<pid>/task`), and `kvm_events_for_request` filters on that set and reports `kvm_events_other_pids`. A run with no recorded tid set attributes nothing rather than guessing. Reds: `Lists differ: [1001, 1002, 2001] != [1001, 1002]`, and for the tid-vs-pid trap specifically, filtering on a pid keeps zero events.

**ftrace overflow.** Now read from every `per_cpu/cpu*/stats` and summed into `ftrace_lost_events`; the analyzer reports `kvm_trace_lossy` and withholds the counts rather than publishing a lower bound. Red: `0 != 49 : cpu0 alone reports zero while cpu3 overran`.

**Serve identity.** `config.uffd_mode` is in the state entry, so `serve_pid_for(tag, umode)` now matches both, defaulting to `copy` when absent (which is what `snapshot serve` defaults to). `start_serve` additionally refuses to adopt any pid it did not spawn. Red: `4242 is not None : a minor-mode serve must not answer for the copy-mode cell`.

**Teardown.** `stop_serve` reaps after `kill()` and then confirms both the pid and the state entry are gone, raising otherwise. Every serve is started up front and stopped in a `finally`, which also removes the leak window you flagged separately.

**Scheduling.** `build_schedule` shuffles the measured requests from a seed recorded in `run.json` and in every record; warmups stay in front so no cell is measured cold. `host_precheck` refuses a host with a foreign Firecracker or load above 1.0, and a `LoadSampler` writes `samples/loadavg.jsonl` throughout. Red: `[file-4k x6, uffd-4k-copy x6] == [...] : the measured order must not be cell-blocked`. Cross-cell interleaving is what made the serves-up-front change necessary.

**guardsupervise.** Sweeps on the normal path too. One thing worth flagging: the obvious implementation is wrong, because `open(cgroup_kill, "w")` CREATES an ordinary file, so a successful write proves nothing. My first version therefore treated the test's temp directory as a real cgroup and polled a static file for the full drain, which turned the reqscale suite from 6s into 31s and broke `test_guardsupervise_terminates_the_third_party_process_group`. It now detects a cgroup by `cgroup.kill` already existing; the suite is 1.1s. Red for the finding itself: `0 != 126 : a cgroup that will not empty must fail the request, not report the child's own status`.

**Report.** `analyze()` was discarding `_validate_schedule`'s return value, so the config was not even in scope. It is now carried into the analysis dict, and the report formats ramp, score, warmup and seed from it, plus a header naming host, kernel, machine, CPU count, fcvm version and sha, source revision (flagged when dirty), Chromium version and a `provenance.json` pointer.

**Smaller ones.** The supervisor handshake is bounded at 10s and the supervisor is killed and reaped in a `finally`. The two `wait_for_stderr_eof` tests are wrapped in `tokio::time::timeout`. The README clause is completed, though not as I first wrote it: I had claimed a second serve costs a second copy of the page cache, then checked and it does not, since both processes mmap the same file. It now states the actual effect, which is that splitting halves what one crashed or overloaded handler takes down.

Gate: `make test-chromium-fault` 23/23, `test_reqscale` 66/66, `test_reqbench` unchanged at 105 with the same 20 failures and 4 errors `origin/main` produces here, `cargo fmt` and `cargo clippy --all-targets` clean, `cargo test --lib` 339 passed with the same 7 uffd failures the box produces with my changes stashed. `faultbench.py` itself was not run end to end: it needs KVM and golden snapshots, which this host does not have, so every rule it now enforces is covered by unit tests instead.

ejc3:
All four fixed in e593f84b. Every one was on the previous commit's own code, and one of them was a fix that made things worse.

**Teardown aborting itself.** You are right that this was backwards: `stop_serve` raising out of the cleanup loop skipped the remaining serves, the load sampler, the host server and the ftrace instance, so a single failed teardown leaked all of them, which is the opposite of what raising was for. Failures are collected, the rest of the teardown runs, and the run fails at the end naming all of them.

**Signal path silent about survivors.** Fixed: both paths report the survivors and exit 126.

**Contradictory message.** Fixed. It now reads `was killed but /proc/<pid> still exists 10s later`.

**Self-enrollment versus a third-party survivor.** This one paid for itself immediately. Seeding a real third-party pid instead of relying on the supervisor's own enrollment exposed that `guardsupervise` opened `cgroup.procs` with `"w"`:

```
AssertionError: '1465767' not found in "guardsupervise: 1 process(es) still in
/tmp/tmpeb42nk1f after cgroup.kill: ['1465768']"
```

The seeded survivor was gone and the supervisor had named itself. Writing to a real `cgroup.procs` MOVES one pid into the cgroup and never replaces its membership, so truncating misdescribes the operation, and against any file-backed stand-in it silently drops every other member. It appends now. Without your distinction the test would have passed on the supervisor's own pid and the truncating write would still be there.

`make test-chromium-fault` 23/23, `test_reqscale` 66/66.

ejc3:
@coderabbitai review

PR #796, opened 2026-08-10
Add fcvm podman prepare with a nameable, keepable snapshot

ejc3:
Adds `fcvm podman prepare`: build a startup snapshot, verify it, reap the VM that produced it, and let the caller name, keep, and rebuild the result.

**Stacked on:** main (817ed25b)

## Problem

Startup snapshots only get created as a side effect of `podman run`, and only when the run was given an explicit `--health-check` URL. To build one deliberately you have to script it: run `podman run` in the background, grep its log for a readiness line, poll `fcvm ls --json` until `health_status` is healthy, run `snapshot create --tag`, kill the VM. Teardown at the end is best-effort, so the script cannot tell whether host resources survived.

There is also no way to check that a snapshot on disk is complete. `check_podman_snapshot` returns `Some` as soon as a `config.json` parses. It never checks that the memory, vmstate and disk artifacts exist, are non-empty, or belong to the generation the config names.

## What it does

`prepare` takes the same arguments as `podman run`.

Cache miss: boot one disposable VM, wait for the health monitor's first Healthy transition, install a full startup snapshot with the source left paused, reap every host resource, print one JSON line.

Cache hit: verify the installed generation and print the same line without booting anything.

```
{"status":"prepared","cache":"created","snapshot_key":"cb-req-golden",
 "content_key":"a1b2c3d4e5f6-startup","snapshot_type":"user",
 "generation_id":"<uuid>","config_digest":"<sha256>"}
```

- Readiness uses the existing health monitor. Image HEALTHCHECK, container-running state and the optional HTTP check keep the authority they have in a run. The only difference is that `prepare` arms the startup trigger unconditionally, where `run` arms it only for an explicit HTTP check.
- `create_snapshot_core` takes a new `SnapshotSourceDisposition`. `run` resumes its VM as before. `prepare` leaves the source paused on every post-pause return, including failure paths, so it cannot mutate past the captured point.
- `cleanup_vm_verified` attempts every teardown action, checks the VMM and namespace holder are gone by `(pid, start_time)` rather than pid alone, and returns an aggregated error. `prepare` does not print its JSON if that fails. `cleanup_nfs_exports` returns its failures for the same reason. `run` keeps best-effort teardown.
- Verification holds a shared lease on the snapshot directory and fsyncs each artifact, the generation directory and the snapshot root before returning. The lease is held until the JSON is written, so a concurrent creator or deleter cannot swap the generation between verification and the answer.
- stdout carries only the JSON line. The output listener still drains the guest so it cannot backpressure, it just does not forward. Logs go to stderr and Firecracker's stdio goes to `/dev/null`.
- The wait for first Healthy is bounded at 600s. `run` needs no equivalent because an unhealthy VM is still a VM the caller owns.

## Naming, keeping, and rebuilding the artifact

`--tag <name>` installs the generation under that name with `SnapshotType::User`, the same contract `snapshot create --tag` writes: the tag is a snapshot directory name, validated by `validate_snapshot_name`, with `config.name` matching. `snapshots ls`, `snapshot serve <tag>`, `snapshot run --snapshot <tag>` and `snapshots delete <tag>` address a prepared tag and a created one identically. Without `--tag` the generation stays at the content-addressed key as `System`.

The prune rule follows from that one choice instead of being a second rule. `snapshots prune` reclaims `System` by default, so an untagged prepare is cache and a tagged one survives until `--all`. The predicate is now one function, `prune_reclaims`.

The tag renames the artifact; it does not replace the cache key. `SnapshotConfig` gains `content_key`, recording the key whose content a generation holds, and `SnapshotConfig::content_key()` falls back to `name`, so every generation already on disk answers the same question without a migration. A `prepare --tag` whose tag holds different content is a miss and rebuilds, exactly as a changed cache key would. A mismatch at the content-addressed key itself is a corrupt generation rather than other content, so it stays an error instead of a ten-minute rebuild that then refuses to publish.

`--force` clears both short circuits: `PreparedTarget::publish_installed` gates the pre-boot check that publishes without booting, and `ExistingGeneration` gates the under-lock check that reuses whatever another process installed. `prepare` never restores from the pre-start cache, so a forced rebuild is a genuine cold boot. It is the only way to rebuild content whose cache key did not change, which is what a repointed remote image reference does.

`CreateSnapshotParams` carries the install name, the content key, the snapshot type and the existing-generation policy. `podman run`'s three call sites take them from `CreateSnapshotParams::cache_entry`, which reproduces the old behaviour by construction: name equals key, `System`, `Reuse`.

## What this unlocks for `bench/chromium`

`reqbench.sh`'s `golden` phase hand-rolls exactly this: background `podman run`, log grep for readiness, `fcvm ls --json` poll, `snapshot create --tag "$TAG"`, kill. It can become one blocking command whose exit status is the answer:

```bash
fcvm podman prepare --name "$TAG-src" --tag "$TAG" ${FORCE:+--force} \
  --network "$EGRESS" --publish 9222:9222 "$IMAGE" > golden.json
```

Four things the benchmark gets that the script could not have:

- `snapshot serve "$TAG"` and `snapshot run --snapshot "$TAG"` keep working unchanged, because a prepared tag is an ordinary snapshot name. The matrix (egress paths x memory backends x page sizes) addresses `$TAG` the same way in every cell.
- The golden snapshot survives `snapshots prune`, so a cleanup between matrix cells cannot silently delete ten minutes of work and turn the next cell into a cold build the report would attribute to the cell.
- A changed image rebuilds by itself, because the content key changes. The benchmark cannot measure a stale golden while believing it measured the new one, which is the class of defect `bench/chromium/AGENTS.md` exists to prevent. `--force` covers the case where the content moved but the reference did not.
- Teardown is verified rather than best-effort. A leftover VM, holder, or pasta from the golden build would sit on the box for the whole measurement run and inflate every number; `prepare` fails loudly instead of printing a durable-looking artifact next to surviving host state.

## Test evidence

`make lint` is fmt + clippy `-D warnings` + audit + deny.

```
$ make lint
cargo fmt -p fcvm -p fuse-pipe -p fc-agent -p failpoint --check
cargo clippy --all-targets -- -D warnings
cargo audit
    Scanning Cargo.lock for vulnerabilities (376 crate dependencies)
warning: 1 allowed warning found          # RUSTSEC-2025-0141 bincode, pre-existing
cargo deny check
EXIT=0
```

clippy printed nothing under `-D warnings`. The audit warning is the allowlisted unmaintained `bincode`. `cargo deny` also prints pre-existing warnings for the `fuser` git source, seven duplicate crate versions, and an unmatched `Zlib` license allowance; this branch changes no `Cargo.toml`, `Cargo.lock` or `deny.toml`.

```
$ make test-unit
     Summary [  50.901s] 639 tests run: 632 passed, 7 failed, 0 skipped
```

The 7 failures need `/dev/userfaultfd`, which this host refuses:

```
thread 'uffd::server::tests::test_handler_failure_kills_the_clones_vmm' panicked at src/uffd/server.rs:
creating userfaultfd (via /dev/userfaultfd):
OpenDevUserfaultfd(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })

$ sysctl vm.unprivileged_userfaultfd
vm.unprivileged_userfaultfd = 0
$ ls -l /dev/userfaultfd
crw------- 1 root root 10, 256 /dev/userfaultfd
```

The same 7 fail on untouched `origin/main` in a separate worktree:

```
$ git worktree add --detach ../wt-mainbase-verify origin/main
$ make test-unit FILTER="-E 'test(/uffd::/)'"
     Summary [   5.477s] 68 tests run: 61 passed, 7 failed, 548 skipped
```

Six `make test-unit` runs on the final tree, two of them after forcing a full rebuild, all reported the same 632/7 and the same seven test names. One earlier run, taken in the minute two sibling worktrees were also building, reported `617 passed, 22 failed`; its per-test output was not captured and it has not recurred. Flagging it rather than explaining it away.

### Red/green

Each gap was reverted on its own and the failure witnessed. The podman+snapshots suite is 27 tests, all green before each revert.

The readiness deadline. Against the unbounded wait the test hangs and nextest terminates it:

```
$ make test-unit FILTER="prepare_gives_up_when_the_container_never_becomes_healthy"
   TRY 2 TMT [ 120.004s] (1/1) fcvm commands::podman::tests::prepare_gives_up_when_the_container_never_becomes_healthy
     Summary [ 245.011s] 1 test run: 0 passed, 1 timed out, 577 skipped
```

The tag. Reverting the install name to the content key:

```
thread 'a_tag_names_the_installed_snapshot_and_no_tag_uses_the_content_key' panicked:
assertion `left == right` failed
  left: "0123456789ab-startup"
 right: "cb-req-golden"
thread 'a_tag_hits_only_for_the_content_it_holds' panicked:
a tag holding this content is a hit
     Summary 27 tests run: 25 passed, 2 failed
```

The type. Reverting a tagged prepare to `SnapshotType::System`:

```
thread 'a_default_prune_keeps_a_tagged_prepare_and_reclaims_an_untagged_one' panicked:
a golden snapshot must not evaporate on the next prune
thread 'a_tag_hits_only_for_the_content_it_holds' panicked:
a tag holding this content is a hit
     Summary 27 tests run: 25 passed, 2 failed
```

`--force`. Reverting `publish_installed` to `true` and dropping `existing` from the under-lock check:

```
thread 'force_rebuilds_instead_of_reusing_an_installed_generation' panicked:
--force must not publish an installed generation (tag None)
thread 'only_a_reuse_policy_keeps_a_generation_installed_by_another_process' panicked:
assertion failed: !snapshot::keeps_installed_generation(ExistingGeneration::Replace, true)
     Summary 2 tests run: 0 passed, 2 failed
```

The cache-key guard. Dropping the `ensure!` so a corrupt generation there becomes a rebuild:

```
thread 'a_mismatch_at_the_content_addressed_key_is_an_error_not_a_rebuild' panicked:
a disk-only generation at the cache key cannot be published
     Summary 1 test run: 0 passed, 1 failed
```

All reverts undone, the affected suite green:

```
$ make test-unit FILTER="-E 'test(/commands::podman::tests::/) or test(/commands::snapshots::tests::/) or test(/cli::args::tests::prepare/) or test(/storage::snapshot::tests::/)'"
     Summary [   0.128s] 49 tests run: 49 passed, 590 skipped
```

`await_prepare_healthy` takes the VM-exit future from its caller so the cancelled, VM-exited, monitor-dropped and deadline outcomes are all reachable without a hypervisor. `prepare_install_target`, `prepared_generation_mismatch`, `keeps_installed_generation` and `prune_reclaims` are pure, and `verify_prepared_snapshot_in` takes its snapshot root as an argument, so the tag, type, content-match and force decisions are all exercised on a tempdir.

## Needs privileged CI

Root VM tests do not run on this host, so `prepare` has never booted a VM. Unproven here:

- Readiness firing against a real container, with an image HEALTHCHECK and with an explicit `--health-check` URL.
- `LeavePaused` holding the source paused through the whole capture, including the diff-tracking retry path.
- `cleanup_vm_verified` succeeding on a real teardown, and returning the aggregated error when a resource survives. This makes `prepare` fail on teardown failures that `run` only logs, so a network cleanup that is flaky under load becomes a command failure.
- The cache-hit path verifying and republishing without booting.
- A tagged generation restoring under `snapshot serve` (UFFD) and `snapshot run --snapshot` (file-backed).
- `--force` reaching the create path and installing over a real generation, and a rebuilt tag leaving no stale artifacts behind.
- The 600s deadline against a slow first boot on a loaded runner.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -25,6 +25,26 @@ The stale `db.sql` disaster happened because setup "usually" cleaned up before s
 
 **Before committing, ask:** "Can any other process/thread/test touch this state at the same time?" If yes, add synchronization or make it atomic. No exceptions.
 
+## WRITE PLAINLY, ENGINEER TO ENGINEER
+
+Commit messages, PR descriptions, code comments, docs: all of it is one engineer telling
+another what changed and why. Say that and stop.
+
+- State the change, the reason, and the evidence. Nothing else belongs there.
+- No selling. Cut "worth having", "the real win", "makes it trustworthy", "elegant",
+  "powerful", "seamless", "comprehensive", "robust".
+- No throat-clearing. First sentence names the change. Skip the scene-setting paragraph.
+- No rhetorical scaffolding: "Two things follow from that", "Four properties make this
+  work", "Here's the thing". Make the point instead of announcing it.
+- No staccato triples ("No traceback. No crash. No exception."). One normal sentence.
+- No em dashes. Comma, parenthesis, or a new sentence.
+- Don't narrate the journey. "I first tried X, then found Y" belongs nowhere. Describe the
+  end state.
+- Don't restate the diff in prose. If the code already says it, delete the sentence.
+- A claim needs a number or a command. "Faster" is not a claim; `2.4s -> 0.3s` is.
+- Read it back as if a colleague sent it to you. If it reads like marketing or an essay,
+  rewrite it.
+
 ## VERIFY PLAN AND RUN NEW TESTS LOCALLY
 
 **Before committing, ALWAYS:**
--- a/README.md
+++ b/README.md
@@ -105,6 +105,29 @@ Snapshot a running VM and restore clones from it. Two modes: UFFD (memory server
 ./fcvm snapshot run --pid <serve_pid> --exec "curl localhost"
 ```
 
+### Building a Snapshot Without Leaving a VM Behind
+
+`podman prepare` takes the same arguments as `podman run`, boots one disposable VM,
+waits for the container to report healthy, installs the startup snapshot, and reaps
+every host resource. It prints one JSON line and exits; there is no VM to clean up.
+
+```bash
+# Build (or verify) the snapshot, then address it by name
+./fcvm podman prepare --name golden --tag nginx-warm nginx:alpine
+{"status":"prepared","cache":"created","snapshot_key":"nginx-warm",
+ "content_key":"a1b2c3d4e5f6-startup","snapshot_type":"user",
+ "generation_id":"...","config_digest":"..."}
+
+./fcvm snapshot serve nginx-warm
+./fcvm snapshot run --pid <serve_pid> --name clone1
+```
+
+The snapshot is keyed by the content of the arguments. Change the image or the VM
+shape and the next `prepare` rebuilds; change the image behind a tag that keeps the
+same reference and use `--force`. Without `--tag` the snapshot is installed under its
+content key as cache, which `snapshots prune` reclaims; with `--tag` it is a user
+snapshot that survives a prune.
+
 ### Disk-Only (Cold-Boot) Clones
 
 `--disk-only` captures just the disk — no vCPU pause, no memory image. The guest
@@ -341,6 +364,7 @@ See [`Containerfile`](Containerfile) for the complete dependency list used in CI
 | `fcvm setup` | Download kernel and create rootfs (5-10 min first run, then cached) |
 | `fcvm setup --cloud-hypervisor` | Build the Cloud Hypervisor backend binary (content-addressed, on demand) |
 | `fcvm podman run` | Run container in a microVM (backend selected with `--hypervisor`, default `firecracker`) |
+| `fcvm podman prepare` | Build and verify a startup snapshot, then reap the VM that produced it (`--tag`, `--force`) |
 | `fcvm exec` | Execute command in running VM/container |
 | `fcvm ls` | List running VMs (`--json` for JSON) |
 | `fcvm snapshot create` | Snapshot a running VM |
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -117,6 +117,8 @@ pub struct PodmanArgs {
 pub enum PodmanCommands {
     /// Run a container in a microVM (Firecracker or Cloud Hypervisor)
     Run(RunArgs),
+    /// Build and verify the healthy startup snapshot, then reap the disposable VM
+    Prepare(RunArgs),
 }
 
 #[derive(Args, Debug)]
@@ -709,9 +711,41 @@ mod tests {
         match cli.cmd {
             Commands::Podman(podman) => match podman.cmd {
                 PodmanCommands::Run(run) => run,
+                PodmanCommands::Prepare(_) => panic!("expected `podman run` command"),
             },
             _ => panic!("expected `podman run` command"),
         }
+    }
+
+    #[test]
+    fn prepare_accepts_the_run_configuration_shape() {
+        let cli = Cli::try_parse_from([
+            "fcvm",
+            "podman",
+            "prepare",
+            "--name",
+            "prepared-web",
+            "--cpu",
+            "4",
+            "--health-check",
+            "http://web.internal/ready",
+            "nginx:alpine",
+        ])
+        .expect("prepare CLI should parse");
+
+        let Commands::Podman(podman) = cli.cmd else {
+            panic!("expected podman command");
+        };
+        let PodmanCommands::Prepare(args) = podman.cmd else {
+            panic!("expected podman prepare command");
+        };
+        assert_eq!(args.name, "prepared-web");
+        assert_eq!(args.cpu, 4);
+        assert_eq!(args.image, "nginx:alpine");
+        assert_eq!(
+            args.health_check.as_deref(),
+            Some("http://web.internal/ready")
+        );
     }
 
     #[test]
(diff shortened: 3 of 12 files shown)


ejc3:
Both findings fixed in 5b21b345, each with a witnessed red first.

**Discard prepared guest output.** The Vec turned out to be dead in every path, not just prepare: both spawn sites (`podman/mod.rs`, `snapshot.rs`) hand it to a `JoinHandle` that is only ever `abort()`ed, and the existing tests abort it too. So a verbose container grew it for the life of *any* VM; prepare is just the worst case, since `emit_output=false` means nothing is even printed while every line is recorded for the full 600s budget. Rather than gate the push on `emit_output`, the listener now forwards and drops, returning `Result<()>`, and the three handle types follow. Red on the accumulating version:

    listener grew 88 MiB while forwarding 150000 lines of 512 bytes; it is
    retaining output that no caller can read (limit 24 MiB)

The test drives the prepare path exactly (`emit_output=false`) and counts `log_tx` deliveries, which fire regardless of `emit_output`, so completion is deterministic rather than timed.

**Verify every snapshot artifact.** Verification now walks `metadata.extra_disks` and each portable volume's `volume-<port>-inode-table.json`, requiring a non-empty regular file, and refuses an extra-disk filename that is not a single normal path component. The inode-table case is the quieter of the two: a missing table does not fail the restore, it renumbers inodes on the clone, which is the glitch the table exists to prevent. Red without the check:

    a missing extra disk must not verify: true

The test also pins the negative: a non-portable volume writes no table, so its absence must still verify.

Gate: `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo audit` clean. `make test-unit` is 641 run / 634 passed; the 7 failures are the `uffd::` tests, which fail identically with these changes stashed (`/dev/userfaultfd` is `crw------- root:root` with `unprivileged_userfaultfd=0` on this host), so they are the box, not the branch.

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

PR #826, opened 2026-08-14
scratch: gate middle-gap probe (auto-closed)

ejc3:
Throwaway PR reproducing the first:N+last:N middle gap.

### DIFF
--- a/MIDGAP_PROBE.md
+++ b/MIDGAP_PROBE.md
@@ -1,0 +1,1 @@
+scratch


MIDGAP_PROBE.md
```diff
@@ -0,0 +1 @@
+scratch
```
ejc3, line 1: P1 probe finding: this silently corrupts output.

ejc3: RED-VERIFIED: scripts/test-check-review-threads.sh :: middle case

ejc3: filler reply posted after the disposition

Issue #841, opened 2026-08-16
Guest podman exec dies with empty conmon output: "container create failed (no logs from conmon): conmon bytes \"\""


ejc3:
## The operation that fails

`podman exec`'s **container create**, inside the guest, during `fcvm exec --pid P -t -- <cmd>` (container path, the default). Everything before it works: the exec handshake completes, `GO` is sent, and the guest returns 177 bytes:

```
Error: container create failed (no logs from conmon): conmon bytes "":
readObjectStart: expect { or n, but found \0, error found in #0 byte of ...||..., bigger context ...||...
reader_loop: received Exit(255)
```

`Exit(255)` is `WIFEXITED` (fc-agent maps signal death to `128 + WTERMSIG`, max 192), so podman exited on its own. The literal does not exist in this repo; it is podman 4.9.3's own text. conmon produced **zero bytes** where podman expected JSON.

## Occurrences

| Run | Job | Head | When |
|---|---|---|---|
| 31889445053 | 95023491433 `Host-Root-arm64-SnapshotDisabled` | `19d78dbb` | 2026-08-15T14:26:12Z |
| 31913833184 | 95082795382 `Host-Root-arm64-SnapshotEnabled` | `adde0a09` (PR #839) | 2026-08-15T23:32:29Z |

Same framing both times: `received 177 bytes (total 177)` then the error then `Exit(255)`.

## It is masked almost everywhere

`.config/nextest.toml` gives `test_fuzz_lifecycle_chaos` `retries = 0` by design; everything else retries once. When this failure lands on any other test it is retried away and the job goes green with the count buried in the summary's `(N flaky)`. On #837 the identical guest failure hit `test_exec_rootless` (`TRY 1 FAIL` -> `TRY 2 PASS`) and the job reported success. It only became visible because it landed on the one test that does not retry.

## Excluded by measurement

- **Not a code change in #839.** Occurred on `19d78dbb`, and `git merge-base --is-ancestor fcea7599 19d78dbb` returns 1. `exec.rs`/`tty.rs` contain no `crate::network` references; the failing VM logged `Snapshot miss`, so the changed readiness path was never called for it.
- **Not the exec path on that machine.** 600 TTY execs (150 iterations x 4 concurrent VMs) on the same runner `i-09aaf9e7e5d22b20a`: 0 failures.
- **Not host memory.** `sar` on that runner at the failure: 127 GB available, 1.73% used, load 0.84 on 32 CPUs, no OOM in `dmesg`.
[...]

PR #842, opened 2026-08-16
Push fork-free guest vitals from fc-agent

ejc3:
**Stacked on:** `runner-loss-detection` (PR #840) → `readiness-attempt-budget` (PR #839)

Closes the diagnostic gap behind #841.

## The problem

`podman exec` died inside the guest with `container create failed (no logs from conmon): conmon bytes ""`, twice, on two different commits. Nothing could say why: `--log-driver=none` discards conmon's stderr and the VM is torn down before anyone looks. Host evidence excluded the host (127 GB available, load 0.84 on 32 CPUs, no OOM), which leaves guest-side candidates that were all unrecorded.

## Fork-free is the point, not a style choice

The obvious design is to pull state from the host at the failure site with `fcvm exec --pid P --vm -- sh -c '...'`. That **cannot work for the cases it exists for**: serving an exec makes fc-agent fork a child (`fc-agent/src/exec.rs:568`), so a guest out of pids or threads fails the collection for the same reason it failed the operation. It also cannot reach a VM whose fcvm already exited, an L2 guest, or a test nextest killed.

Everything in `vitals.rs` is `read`/`readdir`/`statvfs` from an already-running process, so it still answers when the guest can no longer create one.

## What it collects, and what each field can name

| Field | Failure it names |
|---|---|
| `pty/nr` vs `pty/max` | conmon allocates a PTY per `-t` exec |
| `statvfs` of `/run` | rootless podman's runroot is `/run/user/<uid>/containers`; ENOSPC means conmon could not create its socket or exit file and wrote **nothing** — #841's exact signature |
| `oom_kill` from `/proc/vmstat` | monotonic, unlike the 128 KiB printk ring a chatty boot wraps |
| cgroup `pids.current`/`max` | the pids controller rejects forks with nothing the caller sees |
| `MemAvailable`, `Committed_AS` | a 1024 MiB guest failing a fork or mmap silently |

## Emission and routing

A compact line every 10s prefixed `[fcvm-vitals]` (DEBUG → per-VM file, **not** the job log, so it runs for every VM without flooding a runner), and the full block at three in-guest failure sites prefixed `[fc-agent] VITALS` (INFO → also in captured test output).

## Safety

A non-local filesystem is named and skipped rather than probed: `statvfs` on a wedged FUSE mount is uninterruptible and would turn a diagnostic into a second hang. Collection is bounded and reports its own truncation rather than going quiet.

## Tests

6 unit tests, including that no section is ever silently dropped (a missing section must report *unavailable*, since absence and "nothing to report" are the confusion this whole PR exists to end) and that an unmounted path reports a reason instead of zero.

```
$ cargo test -p fc-agent vitals      # 6 passed
$ cargo check -p fc-agent --target x86_64-unknown-linux-musl   # fc-agent ships as musl;
$ cargo check -p fc-agent --target aarch64-unknown-linux-musl  # a host-target check proves nothing
```

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -390,6 +390,29 @@ Host side, captured today:
 | filtered `dmesg` | `dmesg-filtered.log` artifact | host OOM, KVM, page-fault signatures |
 | D-state watchdog | live step log | prints nothing when healthy; an ephemeral runner dies with its artifacts, so the live log is the only channel that survives |
 
+Guest side, pushed by fc-agent (`fc-agent/src/vitals.rs`):
+
+| What | When | Why that field |
+|---|---|---|
+| `MemAvailable`, `Committed_AS` | every 10s, and in full at failure sites | a 1024 MiB guest can fail a fork or an mmap with nothing in dmesg |
+| `pty/nr` vs `pty/max` | same | conmon allocates a PTY per `-t` exec; this is the authoritative free count |
+| `statvfs` of `/run` | same | rootless podman's runroot is `/run/user/<uid>/containers`; ENOSPC there means conmon could not create its socket or exit file and therefore wrote NOTHING, which is issue #841's exact signature |
+| `oom_kill` from `/proc/vmstat` | same | monotonic, cannot wrap, unlike the 128 KiB printk ring |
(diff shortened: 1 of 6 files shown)


ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

PR #843, opened 2026-08-16
Restore the nested reflink, delete the copy fallback, guard sun_path

### LINKED ISSUE #810: test_nested_l2_network_nfs is load-sensitive: 843s timeout under the parallel root suite, 151s pass in isolation
## Evidence

Under the full parallel `make test-root` run (2026-08-13, quiet box, branch pasta-readiness):
- TRY 1 FAIL [842.9s], TRY 2 FAIL [843.0s] — `L2 should complete (missing MARKER_L2_OK). Exit status: ExitStatus(unix_wait_status(0))`
- Earlier run the same night (02:47): TRY 1 FAIL [842.7s], TRY 2 PASS [152.0s]
- Immediately afterwards in isolation (`make test-root FILTER=test_nested_l2_network_nfs`): **PASS [150.8s]**, MARKER_L2_OK present. Sibling `test_nested_l2_nfs` shows the same TRY1-fail/TRY2-pass pattern under load.

Failure telemetry from the failing rep (inside L1, supervising L2):

```
03:44:39 WARN health-monitor: VM no longer healthy, reverting to 5s polling
03:46:10 WARN health-monitor: health check still failing consecutive_failures=10 status=Unknown check_ms=5002
03:48:56 WARN health-monitor: health check still failing consecutive_failures=10 status=Unknown check_ms=20065
03:54:30 WARN health-monitor: health check still failing consecutive_failures=60 status=Unknown check_ms=5010
```

L2 went healthy briefly, then Unknown for 10+ minutes, with individual health checks inside L1 taking up to 20 s — a starvation signature in the family of the NV2 event-loop/timer issues (#630/#660 lineage), not a functional NFS failure (the same cell passes clean when the box is quiet).

Predates today's changes: the 02:47 failure ran a tree without any of the #808/#809/#792 commits.

## Repo rule

"If it passes alone but fails in parallel, find and fix the race condition" — this is that case, filed so the suite result is not silently waved through. The fix likely lives at the same altitude as the #630 timer/event-loop work (L1 vCPU starvation while the host is loaded), not in the test.

## Cheap-repro ideas
- Pin host load while running the single test (e.g. `stress-ng --cpu N`) to reproduce without the 30-minute suite.
- Compare L1 `dd if=/dev/zero of=/dev/null` throughput during the failing window (healthy ≈19 GB/s on Graviton3 per AGENTS.md) to confirm starvation vs. NFS.

ejc3:
**Stacked on:** `guest-vitals` (PR #842) → `runner-loss-detection` (#840) → `readiness-attempt-budget` (#839)

Fixes #810. It is a regression from `55ef6350` (2026-01-04), not a load problem — reflinks did work before it.

## The chain, from that commit's own message

1. A prior commit set `FCVM_DATA_DIR=/root/fcvm-data` for nested VMs "to avoid Unix socket issues on FUSE filesystems".
2. That left the base image on the mapped mount and the VM disk on the L1's local disk, so `FICLONE` became cross-superblock. The VFS refuses that with `EXDEV` **before FUSE is ever consulted**.
3. The same commit added a `cp` fallback, "slower but works", measured at 327 s on a quiet box.

An O(1) clone had become a ~10 GB copy pulled through fuse-pipe, per nested VM, and the only trace was one `WARN`.

Under the parallel suite that copy exceeds the `timeout 600` bounding the L2 launch, so `MARKER_L2_OK` never appears. Four failures inside a **0.4 s band** across three tests is a deadline, not a stall:

```
842.664s  test_nested_l2_network_fuse  TRY 1 FAIL
842.696s  test_nested_l2_nfs           TRY 1 FAIL
842.936s  test_nested_l2_network_nfs   TRY 1 FAIL
843.033s  test_nested_l2_network_nfs   TRY 2 FAIL   <- failed both tries
```

## Measured before changing anything

- **AF_UNIX bind and connect both succeed on the fuse-pipe mount**, tested in a real guest with a local-disk control. The stated premise does not hold.
- fuse-pipe has had `mknod` for special files including sockets since `26c75963` (2025-11-28), **five weeks before** that workaround.
- The L1's local disk is **ext4**, which has no reflink at all — so there was never a local arrangement in which CoW could have worked.

## What January most likely hit: path length, not the filesystem

`sun_path` is 108 bytes. Reproduced here: a uuid-keyed data dir gives a **128-byte** socket path, `bind` fails, and it surfaces as `VolumeServer 0 failed to signal ready: channel closed` — naming neither the path nor the limit.

## The change

- Data dir moves back onto the mapped mount at a **short** path (`/mnt/fcvm-btrfs/nd/<8 hex>`, 91 bytes for the longest socket). A uuid rather than `$$`, because concurrent L1s share that directory and separate PID namespaces reuse numbers.
- **The cross-device `cp` fallback is deleted.** A reflink that silently becomes a copy is a different operation, not a slower success. Failure now names which of the two causes applies, since the fix differs per cause.
- `paths::check_socket_path_budget` refuses an over-deep data dir *before* the VM directory is created, naming the path and the limit.

## Verified end to end (aarch64)

| | before | after |
|---|---|---|
| `test_nested_l2_fuse` | `TRY 1 FAIL [901.890s]` | `PASS [173.927s]` |
| cross-device fallbacks | 1 | **0** |
| markers | — | `MARKER_L1_OK`, `MARKER_L2_OK` |

## Tests

4 for the `sun_path` budget, including the exact 128-byte path that broke it and an exact-boundary case (at the limit passes, one byte over fails), plus an assertion in the nested test that the L1 log carries no cross-device fallback. That assertion is red against tonight's captured log.

Note for review: the nested tests are excluded from CI by `Makefile` `CI_ROOT_BASE`, so this verification is local and CI will not exercise it.

### DIFF
--- a/src/commands/podman/mod.rs
+++ b/src/commands/podman/mod.rs
@@ -1353,6 +1353,13 @@ async fn prepare_vm_for_lifecycle(
     }
 
     // Setup paths
+    //
+    // Checked before anything is created: the VM's Unix sockets live in this
+    // directory, and a directory a few characters too deep pushes them past
+    // sun_path. bind() then fails with ENAMETOOLONG and surfaces far away from
+    // the cause (observed as "VolumeServer failed to signal ready: channel
+    // closed"), so refuse it here where the path can still be named.
+    paths::check_socket_path_budget(&vm_id)?;
     let data_dir = paths::vm_runtime_dir(&vm_id);
     tokio::fs::create_dir_all(&data_dir)
         .await
--- a/src/paths.rs
+++ b/src/paths.rs
@@ -132,7 +132,118 @@ pub fn vm_runtime_dir(vm_id: &str) -> PathBuf {
     data_dir().join("vm-disks").join(vm_id)
 }
 
+/// `sockaddr_un.sun_path` is 108 bytes including the NUL, so a path of 107
+/// bytes is the most that can ever be bound.
+const SUN_PATH_MAX: usize = 107;
+
+/// The longest socket file placed in a VM's runtime directory.
+///
+/// Checked rather than assumed: if a longer name is added later, the budget
+/// below silently stops covering it, so this constant and the check move
+/// together.
+const LONGEST_SOCKET_NAME: &str = "firecracker.socket";
+
+/// Fail early when this VM's data directory leaves no room for its sockets.
+///
+/// A data directory only a few characters too deep pushes the VM's sockets past
+/// `sun_path`, `bind` returns ENAMETOOLONG, and what the operator sees is a
+/// downstream symptom with no mention of paths at all. Observed 2026-08-16: a
+/// nested data directory keyed by a full uuid produced a 128 byte socket path,
+/// and the failure surfaced as
+///   "VolumeServer 0 failed to signal ready: channel closed"
+/// which names neither the path nor the limit. Worse, the same class of failure
+/// appears to have been misdiagnosed in 55ef6350 as "FUSE doesn't support Unix
+/// sockets", and the workaround for that non-problem disabled reflinks for
+/// every nested VM for seven months (issue #810).
+pub fn check_socket_path_budget(vm_id: &str) -> anyhow::Result<()> {
+    check_socket_path_budget_under(&data_dir(), vm_id)
+}
+
+/// [`check_socket_path_budget`] against an explicit data directory.
+///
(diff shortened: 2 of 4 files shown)


ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

PR #848, opened 2026-08-16
Record what a request cost in memory, on the record that has its latency

ejc3:
The request path records **no memory at all**. The only memory-ish field in a run's `analysis.json` is `cell.memory_mib` — the guest's *configured* RAM, not a measurement. So nothing answers "how much did this clone dirty between restore and screenshot".

The only memory numbers in the project come from a separate concurrency ladder run by `bench.sh`. A latency-versus-memory frontier chart built from those would be a **join across two harnesses, two runs and two goldens**, where every point silently crosses a run boundary. The report already carries that defect in its cross-run subtractions; a new chart shouldn't add another.

## The change

`teardown_fast` already pins the exact child set and samples per-child CPU immediately before the kill. This samples `Private_Dirty` at the same point and publishes `per_child_memory` beside `per_child_cpu`.

**Why there and nowhere else:** a clone starts as a view of the shared snapshot, so `Private_Dirty` is what it *privatised* — pages written rather than read — which is precisely the delta the request cost. It must be read while the process is alive: `smaps_rollup` dies with the address space, so unlike CPU it cannot be recovered from a zombie. The sample also lands before `t_kill`, so it cannot perturb the reap timing the same function measures.

## It never reports a confident zero

The sampler returns a **reason** rather than a zero when it cannot measure (exited, no `CONFIG_PROC_PAGE_MONITOR`, permission), and the total is `None` rather than `0` when nothing could be sampled. A zero with no uncertainty is a claim, and a failed read does not support one.

```
self:          {'rss_kb': 22916, 'pss_kb': 15105, 'private_dirty_kb': 13416}
dead pid:      {'private_dirty_kb': None, 'unavailable': 'no smaps_rollup (exited, or kernel lacks CONFIG_PROC_PAGE_MONITOR)'}
```

## Tests

`PerRequestPrivateDirty`, 3 cases: a live process reports a positive figure; an unreadable one reports a reason and `None`; a run where nothing could be sampled totals `None` rather than a confident 0 KiB.

```
$ python3 -m unittest discover -s bench/chromium
272 passed
```

Enables the frontier chart to be one measurement: latency and delta-memory from the same golden, same seal, same record.

## Summary by CodeRabbit

* **New Features**
  * Added per-process private memory usage sampling during benchmark teardown.
  * Reports total private dirty memory when measurements are available.
  * Identifies processes whose memory usage could not be measured and provides reasons.

* **Tests**
  * Added regression coverage for successful sampling, unavailable process data, and missing measurement totals.

### DIFF
--- a/bench/chromium/reqbench.py
+++ b/bench/chromium/reqbench.py
@@ -136,6 +136,42 @@ CONTROL_WINDOW_S = 0.05
 # ---------------------------------------------------------------- procfs utils
 
 
+def proc_private_dirty_kb(pid: int) -> dict:
+    """What this process has PRIVATISED, from /proc/<pid>/smaps_rollup.
+
+    A clone starts as a view of the shared snapshot, so its Private_Dirty IS
+    the delta it cost: the pages it wrote rather than read. Read here, alive,
+    immediately before the kill, so the figure is the cost of the request that
+    just completed.
+
+    Not readable from a zombie. smaps_rollup disappears with the address space,
+    which is why this cannot be sampled alongside the CPU figures after exit.
+
+    Returns the reason rather than a zero when it cannot measure. A zero with no
+    uncertainty is a claim, and an unreadable file does not support one.
+    """
+    try:
+        with open(f"/proc/{pid}/smaps_rollup", "r", encoding="ascii") as handle:
+            rollup = handle.read()
+    except FileNotFoundError:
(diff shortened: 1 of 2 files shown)


ejc3:
Superseded: both commits landed on main as edcad52f and f167c726 (same content, re-pushed), and 58ba1c71 then corrected the teardown_fast comment this branch still carries ('The delta this request cost' vs main's 'an absolute reading, NOT a delta — no baseline is subtracted anywhere'). Merging this branch now would conflict there and restore the retracted mischaracterization, so closing rather than landing; nothing in it is unlanded.

PR #850, opened 2026-08-16
Name podman's store explicitly when exporting via skopeo

ejc3:
**Stacked on:** `restore-record` (PR #849)

skopeo resolves `containers-storage:` from storage.conf alone and applies no rootless remap, so on a host whose `/etc/containers/storage.conf` pins a root-only runroot (observed: `runroot = "/run/containers/storage"`, written by the host's config management), a rootless export of a localhost image fails with `mkdir /run/containers/storage: permission denied` for an image rootless podman built moments earlier.

The export now asks podman for the store it actually uses (one `podman info` call) and names it as an explicit `[driver@graphroot+runroot]` specifier. On any failure to resolve, the specifier is empty and the bare transport behaves exactly as before, so hosts that never had the problem see no change.

## Test Results

```
$ cargo test -p fcvm --lib store_specifier
2 passed
```

On the affected host, the same rootless command red/green:

```
# before
Error: skopeo copy failed for image 'localhost/www-demo-selftest' (...): mkdir /run/containers/storage: permission denied
# after
$ fcvm podman prepare --name selftest --tag demo-selftest3 --publish 18080:80 localhost/www-demo-selftest
{"status":"prepared","cache":"created","snapshot_key":"demo-selftest3",...}   # rc=0
```

### DIFF
--- a/src/commands/podman/image.rs
+++ b/src/commands/podman/image.rs
@@ -176,7 +176,11 @@ pub async fn export_image_archive(image_id: &str, repo_tag: &str, dest: &Path) -
     let output = tokio::process::Command::new("skopeo")
         .args([
             "copy",
-            &format!("containers-storage:{}", image_id),
+            &format!(
+                "containers-storage:{}{}",
+                podman_store_specifier().await,
+                image_id
+            ),
             &format!("docker-archive:{}:{}", dest_str, repo_tag),
         ])
         .output()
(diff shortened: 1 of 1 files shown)


ejc3 (commented):
NOT-A-DEFECT: the CodeRabbit summary comment is a walkthrough, not a finding; it raised no inline findings on this PR.

NOT-A-DEFECT: the chatgpt-codex-connector comment says the reviewer hit its usage limit and never ran, so it carries no finding. A local GPT-5.5 xhigh adversarial pass reviewed this commit instead; its one finding here (multi-line podman-info output could embed a newline in the store specifier) is fixed in 3ef480cf with a unit test watched failing first.

PR #851, opened 2026-08-16
Carry pasta's own error into network-setup failures

ejc3:
**Stacked on:** `skopeo-rootless-store` (PR #850)

pasta creates its TAP and writes its readiness PID file before binding forwarded ports, so a port conflict kills it after every readiness signal fcvm watches. The next setup step then failed with `bridge setup failed: ... Cannot find device "pasta0"` while the only line naming the cause (`Listen failed for HOST TCP port ...: Address already in use`) was dropped twice over: the propagated error never consulted the dead child, and the stderr stream was logged under a bare `pasta` target that every documented `RUST_LOG=fcvm=...` invocation filters out.

Two changes:
- bridge-setup failures append pasta's exit status and captured stderr tail when pasta has exited;
- the stderr stream logs under `fcvm::pasta`, so the standard filters keep it (in per-VM debug logs and CI artifacts too).

A wildcard `0.0.0.0` listener collides with a pasta forward on any per-VM loopback IP, which is how one host daemon squatting one port broke every `--publish` VM on a shared dev box, with three consecutive runs blaming the bridge.

## Test Results

`tests/test_pasta_diagnostics.rs` squats a wildcard port (kernel-assigned, so parallel-safe) and asserts the run fails with pasta's own error in the output. Watched fail without the fix and pass with it:

```
# fix reverted (git stash), rebuilt:
thread 'port_conflict_failure_names_pastas_own_error' panicked:
the failure must carry pasta's own stderr (the port conflict), not only a downstream symptom
test result: FAILED. 0 passed; 1 failed
# fix restored:
test result: ok. 1 passed; 0 failed (0.77s)
```

### DIFF
--- a/src/network/pasta.rs
+++ b/src/network/pasta.rs
@@ -64,6 +64,13 @@ const PASTA_STDERR_TAIL_LINES: usize = 20;
 /// case of an inherited, still-open pipe.
 const PASTA_STDERR_EOF_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(2);
 
+/// Bounded window for a dying pasta to become reapable when a network-setup
+/// step fails. pasta drops its TAP before it can be reaped, so the failing
+/// step routinely observes the vanished device a beat before `try_wait`
+/// reports the exit; sampling once would drop the diagnosis exactly when it
+/// is needed.
+const PASTA_EXIT_CONTEXT_WAIT: std::time::Duration = std::time::Duration::from_secs(2);
+
 /// Whether `ip neigh show` proves that the bridge resolved the guest's MAC.
 ///
 /// This is an L2 fact and nothing more. A resolved entry is necessary for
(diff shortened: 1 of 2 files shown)


ejc3 (commented):

NOT-A-DEFECT: the chatgpt-codex-connector comment says the reviewer hit its usage limit and never ran, so it carries no finding. A local GPT-5.5 xhigh adversarial pass reviewed this commit instead; its two findings here are fixed in ec32c4cd: pasta_exit_context now waits a bounded window for a dying pasta and reports captured stderr even when it is still running, and the integration test goes through the common spawn helper (pdeathsig, log consumers, bounded wait) and is gated behind privileged-tests like every VM-booting test. The Container-x64 failure of the ungated test was watched red for exactly the assets reason the gate comment states; the one Host-Root-x64 fuzz failure carried its own attribution (guest answered TCP, all five neighbour queries stalled, host load 119) and did not reproduce on the regated run.

PR #853, opened 2026-08-16
Route bench markdown to bench-tests instead of nothing

ejc3:
## The hole

`case` globs match `/`. The `changes` job classified paths with the docs arm first:

```sh
case "$f" in
  scripts/claude-assistant/*|.github/workflows/claude*.yml|.claude/*|docs/*|*.md) ;;
  bench/*) bench=true ;;
  *) code=true ;;
esac
```

so `bench/chromium/README.md` matched `*.md` before `bench/*` was ever considered. That PR classifies as `code=false bench=false`: every gating job skips and `Summary` renders clean.

`bench/chromium/test_reqbench.py` carries four lints that read exactly those files:

| lint | reads |
|---|---|
| `test_every_binomial_bound_matches_reqanalyze_clopper_pearson` | `REVIEW.md` |
| `test_agents_md_jpeg_figures_match_the_record_run` | `AGENTS.md` |
| `test_every_path_cited_in_a_doc_table_is_committed` | `AGENTS.md`, `REVIEW.md` |
| `test_the_readme_healthcheck_verification_actually_fails` | `README.md` |

So the one class of change those lints exist to catch was the one class that ran nothing. That is the green-by-absence pattern `ci.yml` already documents twice in its own comments.

Found while correcting `bench/chromium/README.md` and `REVIEW.md` in #847 — a PR editing only those files would have run zero jobs.

## Why it was easy to miss

The `*.md` in `paths-ignore:` at the top of the same file is a **different matcher**. GitHub's filter globs stop at a slash, so there it means root-level markdown only. Identical spelling, opposite behaviour. That contrast is now a comment.

## The fix

`bench/*` precedes `docs/*|*.md`.

## Tests

Both lift the **shipped** `case` block out of `ci.yml` and run it under `bash`, rather than restating the globs. The defect is an ordering bug between two arms that are individually correct, so a restatement would carry the author's assumed ordering and pass.

- `a_bench_markdown_change_routes_to_the_bench_tests` — the four bench doc files
- `path_classification_holds_on_both_sides_of_the_bench_arm` — ten paths pinned on both sides, so a reorder cannot trade this hole for another

### Red-verified against the unfixed tree

```
$ make _test-unit FILTER="-E 'binary(test_ci_workflow_coverage)'"
TRY 2 FAIL a_bench_markdown_change_routes_to_the_bench_tests
  `bench/chromium/README.md` classified as code=false bench=false: it matches the
  docs arm before `bench/*`, so a PR touching only it gets zero gating jobs while
  bench lints that read it exist
Summary 19 tests run: 18 passed, 1 failed
```

With the fix, and again red after reverting **only** `ci.yml`:

```
Summary [0.028s] 19 tests run: 19 passed, 0 skipped
```

The first draft of the harness read the leading two words of stdout, which made `ci.yml`'s own `echo "bench path: $f"` parse as the flags and reported correctly-classified bench sources as defects. It keys off a `CLASSIFIED` marker line instead — caught precisely because the sibling test was expected to pass and did not.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CI change detection for non-pull-request events, empty change lists, and renamed files.
  - Correctly classifies benchmark, code, documentation, README, performance, and Makefile changes.
  - Continues excluding unrelated documentation and assistant files from applicable checks.
  - Improves reliability of automated checks by ensuring relevant validation runs consistently across change types.

- **Tests**
  - Added comprehensive coverage for CI path classification, workflow gating, rename handling, and output wiring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,15 +76,31 @@ jobs:
           set -euo pipefail
           # Only a pull_request has a well-defined changed-file set here; every
           # other trigger runs the full matrix rather than guess.
+          # A fail-open must open EVERY gate. Writing `code=true` alone left
+          # `bench` unset, and bench-tests is gated on it, so the bench lints
+          # were skipped on push-to-main, workflow_dispatch and the Build
+          # Kernels workflow_run — i.e. never ran on main at all, which matters
+          # because the documented stacked-PR routine force-merges with --admin
+          # after cancelling the PR-side run. Summary counts skipped as
+          # non-failure, so that read as green.
           if [ "$EVENT" != "pull_request" ]; then
             echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "bench=true" >> "$GITHUB_OUTPUT"
             echo "event $EVENT: running the full matrix"
             exit 0
           fi
-          files=$(gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '.[].filename')
+          # `.filename` is the POST-rename path only. `git mv src/foo.rs
+          # bench/chromium/foo.rs` would report one bench/ entry, leave code
+          # false, skip the whole VM matrix, and make a non-compiling tree
+          # mergeable behind a green Summary. `previous_filename` is present on
+          # exactly the renamed entries, so classifying both paths means a
+          # rename is judged by where it came from as well as where it went.
+          files=$(gh api "repos/$REPO/pulls/$PR/files" --paginate \
+            --jq '.[] | .filename, (.previous_filename // empty)')
           if [ -z "$files" ]; then
-            # An empty file list is unexpected, so fail toward running the matrix.
+            # An empty file list is unexpected, so fail toward running everything.
             echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "bench=true" >> "$GITHUB_OUTPUT"
             echo "no changed files reported: running the full matrix"
             exit 0
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,9 +98,23 @@ jobs:
           code=false
           bench=false
           while IFS= read -r f; do
+            # `bench/*` MUST precede the docs arm. A shell `case` glob matches
+            # `/`, so `*.md` swallows bench/chromium/README.md and that PR
+            # reports code=false bench=false: zero gating jobs, clean Summary.
+            # Meanwhile bench/chromium/test_reqbench.py carries lints that read
+            # exactly those files (the Clopper-Pearson bound lint on REVIEW.md,
+            # the JPEG figure and cited-path lints on AGENTS.md, the healthcheck
+            # lint on README.md), so the one class of change those lints exist
+            # to catch would be the one class that runs nothing.
+            #
+            # The `*.md` in `paths-ignore:` at the top of this file is NOT the
+            # same pattern: GitHub's filter globs stop at a slash, so there it
+            # means root-level markdown only. Identical spelling, different
+            # matcher — which is why this was easy to miss.
             case "$f" in
-              scripts/claude-assistant/*|.github/workflows/claude*.yml|.claude/*|docs/*|*.md) ;;
+              scripts/claude-assistant/*|.github/workflows/claude*.yml|.claude/*) ;;
               bench/*) bench=true; echo "bench path: $f" ;;
+              docs/*|*.md) ;;
               *) code=true; echo "code path: $f" ;;
             esac
           done <<< "$files"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,6 +129,20 @@ jobs:
             # matcher — which is why this was easy to miss.
             case "$f" in
               scripts/claude-assistant/*|.github/workflows/claude*.yml|.claude/*) ;;
+              # Docs that TESTS READ are code, whatever their extension.
+              # tests/test_documented_make_targets.rs reads README.md and
+              # PERFORMANCE.md and fails when they name make targets that do not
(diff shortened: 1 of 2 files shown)


ejc3:
REVIEW-ACK: adversarial /code-review pass at high effort. Two findings, both accepted and fixed in 4458936e.

RED-VERIFIED: a_fail_open_opens_the_bench_gate_too — the fail-open exits wrote `code=true` and never wrote `bench`, so bench-tests was skipped on push-to-main, workflow_dispatch, workflow_run and the empty-file-list path. Watched failing against the unfixed tree for all four events.

RED-VERIFIED: renames_are_classified_by_their_source_path_too — `--jq .[].filename` reports only the post-rename path, so `git mv src/foo.rs bench/chromium/foo.rs` skipped the whole VM matrix behind a green Summary. Watched failing; the query now emits `previous_filename` too, verified against a real renamed-entry payload.

The review also independently reproduced the original fix by extracting the shipped case block and running 18 paths through bash, and ran the suite on this PR head (19/21 at the time, 21/21 now). Its conclusion that nothing which previously routed to `code` changed matches my own pinning test.

ejc3 (commented):
RED-VERIFIED: tests/test_ci_workflow_coverage.rs — a_fail_open_opens_the_bench_gate_too, renames_are_classified_by_their_source_path_too

Disposes all three preceding bodies.

1. codex (COMMENT) and coderabbitai (COMMENT) are rate-limit notices carrying no finding. NOT-A-DEFECT on their content; recording it explicitly because the gate cannot tell a notice from a finding, and that bias is correct.

2. My own earlier body used REVIEW-ACK:, which this gate deliberately no longer accepts — it says so at line 399: a bare acknowledgement meant "an identical claim was adjudicated or waved through depending only on where the reviewer happened to click." That body was a claim, not an answer. This one answers it.

The two real findings came from an adversarial /code-review pass and are fixed in 4458936e, each closed by a test watched failing against the unfixed tree:

- Fail-open opened only the code gate, leaving bench unset, so bench-tests was skipped on push-to-main, workflow_dispatch, workflow_run and the empty-file-list path. Red across all four events.
- `--jq .[].filename` classified a rename only by its destination, so `git mv src/foo.rs bench/chromium/foo.rs` skipped the whole VM matrix behind a green Summary. Now emits previous_filename, verified against a real renamed-entry payload.

21/21 in binary(test_ci_workflow_coverage).

ejc3:
## The x64 `test_fuzz_lifecycle_chaos` failure is not from this PR, and #846 already fixes it

This PR changes `ci.yml`'s path classification and one test file, so I checked rather than assumed — main is green, so "unrelated" needed evidence.

**It is not flaky and it is not contention-as-excuse.** From the clone's own debug log (`fuzz1-snap-3401438-0-clone-5-…log`, artifact `test-logs-host-root-x64-SnapshotEnabled`):

```
guest 10.0.2.100 never appeared at L2 within 5s: the guest ANSWERED TCP but its
neighbour entry never resolved; neighbour: "" (never read: every neighbour query
stalled); probe: answered (SYN-ACK or RST); rounds 5 (5 stalled);
host load 155.57/56.96/27.93, runnable 8/7618
```

Every round: `answered=true neighbor_resolved=false`. The guest was up and answering TCP the entire time. What failed was `nsenter … ip neigh show`, which stalled on all five attempts under a load average of 155 with 7618 threads. The forensics dump taken 1.6s later shows the entry was there and healthy all along:

```
10.0.2.100 dev br0 lladdr 02:4e:29:54:31:ac REACHABLE
```

The same log shows a sibling `nsenter` failing outright (`exit status: 126`), which is the same cause.

So the readiness check failed the VM for the HOST's inability to run `ip neigh` within the budget, not for anything about the guest. #846 fixes exactly this in `src/network/pasta.rs`: when the probe answered and **every** neighbour query across **every** round stalled, it accepts on the TCP evidence and warns that the host was too loaded, rather than tearing down a working VM. The condition is deliberately narrow — a neighbour entry that was read and found unresolved still fails, and a guest that never answered still fails, including the resolved-MAC-but-silent case that check exists to catch.

Reproduced seed 1 locally on arm64: PASSED in 23.5s, versus 87.2s to reach op 5 in CI. Not a schedule difference; a load difference.

Merge order is therefore #846 first, then rebase this PR. Its `Bench Harness Tests` failure has the same answer — it is the `machine=0.000000ms` bug that only #846 fixes.

ejc3 (commented):
NOT-A-DEFECT: the earlier fuzz-failure comment above is a diagnosis, not an open finding — the x64 test_fuzz_lifecycle_chaos failure it analyzes was fixed in #846 (merged: the accept-on-TCP-evidence branch plus the port-forward wait getting its own budget), and this branch has since rebased onto that main and gone green on the same job.

Since then this PR also absorbed its own review round: the changes-job output wiring test (a one-character typo in `outputs:` silently disables every gating job; red-verified against three mutations), the summary gate's `exit 1` body check, the constant-boolean poison check on gated `if:` conditions, the comment-stripping fix for the checkout-ownership guard, and clippy's no_effect_replace on my own dead code. Each defect claim is closed by a test watched failing under its mutation; evidence in the commit messages.

ejc3:
@coderabbitai review

ejc3 (commented):
Reviewed at 95ef533c via the full-stack adversarial pass across all open PRs (code-review + verification): the shipped classifier was executed against 19 probe paths and only the intended classifications changed; the branch's test suites were run against a main control with identical (environmental-only) failure sets. No findings on this PR. NOT-A-DEFECT: no defects to disposition.

ejc3:
@coderabbitai review

ejc3:
@codex review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

PR #855, opened 2026-08-16
Accept delegated subnets of any length in --ipv6-prefix

ejc3:
**Stacked on:** `pasta-error-visibility` (PR #851)

Routed mode required a bare 4-group /64 prefix, but a shared dev host's fabric often routes it a longer delegated subnet (a /112 service subnet in the run that hit this), which the /64-only form could not express: preflight rejected the host's own subnet outright.

```
Error: routed mode preflight check failed: invalid --ipv6-prefix
'2803:6086:9892:947b:1feb:1cca:0002/112': expected 4 colon-separated hex groups
```

`--ipv6-prefix` now also takes standard CIDR notation up to /120 (at least 8 host bits stay available for VM addressing). The VM address hash fills exactly the host bits below the prefix length, suffixes 0 and 1 are reserved (subnet-router anycast, and the host side of the prefix), and set host bits below the length are rejected rather than silently masked — `2600::1/64` is a confused caller, not a subnet. The bare 4-group shorthand keeps meaning /64, so existing invocations are untouched.

## Test Results

```
$ cargo test -p fcvm --lib -- routed
11 passed; clippy clean
```

Covers: CIDR parse for a full-8-group /112 and a compressed ::/64, length bounds (/0, /121), host-bits rejection, the 7-group non-address form failing with a message naming the address part, all prior shorthand behaviors, and `vm_addresses_stay_inside_a_delegated_112` (generated addresses keep the top 112 bits and never land on suffix 0/1).

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -262,11 +262,13 @@ pub struct RunArgs {
     #[arg(long, value_enum, default_value_t = Hypervisor::Firecracker)]
     pub hypervisor: Hypervisor,
 
-    /// Routable IPv6 /64 prefix for routed mode VM addressing.
-    /// Each VM gets a unique address in this prefix via NDP proxy.
-    /// When set, MASQUERADE is skipped (the prefix is directly routable).
-    /// When not set, auto-detected from host interfaces.
-    /// Example: --ipv6-prefix 2803:6084:7058:46f6
+    /// Routable IPv6 prefix for routed mode VM addressing: CIDR notation for
+    /// a delegated subnet of any length up to /120 (e.g.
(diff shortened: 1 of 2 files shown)


ejc3 (commented):

NOT-A-DEFECT: the chatgpt-codex-connector comment says the reviewer hit its usage limit and never ran, so it carries no finding. The change is covered by 11 unit tests (CIDR parse including the motivating /112, bounds and host-bits rejection, shorthand back-compat, and vm_addresses_stay_inside_a_delegated_112), and the motivating live failure and its fix were verified end to end on the host that hit it.

PR #856, opened 2026-08-16
Make prepare's health deadline a flag (--ready-timeout)

ejc3:
**Stacked on:** `routed-prefix-cidr` (PR #855)

`prepare` gave a disposable VM a fixed 600 seconds to report a healthy container. A workload that imports a large image and warms caches before its health check can pass does not fit: a 36 GB image took over 14 minutes just to import in the run that hit this, so prepare tore the VM down mid-warmup every time and the snapshot could never be built:

```
Error: disposable prepare VM did not report a healthy container within 600s
```

`--ready-timeout <secs>` (default 600, the previous constant) bounds the wait, and the deadline error now names the flag. The value changes how one build runs, not what it produces, so it stays out of the snapshot's content key.

## Test Results

```
$ cargo test -p fcvm --lib -- prepare_health prepare_gives_up honors_a_custom
3 passed
```

The existing deadline and non-healthy-outcome tests, plus a new one asserting a 3600s budget is honored and named in the error.

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -160,6 +160,13 @@ pub struct PrepareArgs {
     #[arg(long)]
     pub force: bool,
 
+    /// Seconds to wait for the disposable VM's container to become healthy
+    /// before giving up. The default fits images that start serving quickly;
+    /// workloads that import a large image or warm caches before their health
+    /// check passes need more. Not part of the snapshot's cache key.
+    #[arg(long, default_value_t = 600)]
+    pub ready_timeout: u64,
+
     #[command(flatten)]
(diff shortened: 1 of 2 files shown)


ejc3 (commented):

NOT-A-DEFECT: the chatgpt-codex-connector comment says the reviewer hit its usage limit and never ran, so it carries no finding. The change is covered by the deadline and non-healthy-outcome unit tests plus a new one asserting a custom 3600s budget is honored and named in the error, and the motivating live failure (a 36 GB image import outliving the fixed 600s budget) was verified fixed end to end.

PR #918, opened 2026-09-08
Use the upstream pasta addr_seen fix without local patches

### LINKED ISSUE #661: test_clone_port_forward_stress_rootless: first clone's pre-storm curl RSTs (health-ready vs proxy-path-ready race; reproduces on main)

ejc3:
Pin both pasta build paths to upstream 3f57f0382f6a72c0b8ce0c5ff92248b5117ed9b6 and remove the two carried addr_seen patches. The upstream commit has the same stable patch-id as the runtime patch, 340576aef02fa19411e69e3587566f3c4950057a. This is the existing fix for #661, now supplied by upstream.

Stack: #915 -> #916 -> #917 -> this PR.

The runner tarball is verified with SHA-256 2d698e3f7a96408231aa11bb1b27775de6964e2de7b0fa29399847d012044e3a. Cache locking, atomic publication and the distro fallback are unchanged. No other passt upgrade is included.

Validation:
- make lint passed.
- 17 focused pin and AMI-hash tests passed.
- pasta_build_paths_pin_the_upstream_addr_seen_fix failed on the original pins, passed with this change, failed after a production-only runner-pin reversion, and passed after restoration.
- make setup-passt verified the archive and built/installed the pinned source without patches on ARM64.
- make test-root selected the existing rootless clone port-forward stress test and exited 0: 33.159s, all three first requests and all 60 concurrent requests succeeded, zero empty responses or errors. It selected the newly built pasta-56f1f1608297.bin. All test-owned VM/server processes were reaped.
- Main-agent review of the complete 16-file delta found no blocking issue.

Upstream: https://passt.top/passt/commit/?id=3f57f0382f6a72c0b8ce0c5ff92248b5117ed9b6

Closes #661.


ejc3:
@coderabbitai review

PR #923, opened 2026-09-13
Run one x64 job on pull requests; main runs the rest

ejc3:
## Why
Every PR push launched four x64 metal runners: Host-x64, both Host-Root-x64 modes and Container-x64. c5d.metal is the largest runner line item (about $925 of roughly $2,000 in runner spend since 2026-07-22). Three of those four jobs run paths the arm64 jobs already cover on the same PR.

## What changes
- `skip-check` now plans the self-hosted matrix.
  - **pull_request:** every arm64 job, plus `Host-Root-x64-SnapshotEnabled`. That job takes the privileged suite through a snapshot miss and a restore on x86, which is where the arch-specific KVM and vCPU state code lives.
  - **push to main:** every job that has not already succeeded on a PR run of the same tree. After a normal merge that means the other three x64 jobs, so x86 still gets the full matrix before release. If the PR run skipped the matrix (docs-only), main skips it too, as before.
  - **Any other event:** the full matrix. That covers workflow_dispatch (and so ci-train), the Build Kernels workflow_run, and a push whose tree never passed on a PR.
- `host`, `host-root` and `container` take their matrix from the plan via `fromJSON`, and skip when their list is empty.

## Tests
- New tests in `tests/test_ci_workflow_coverage.rs` run the real planner step with fake `gh` and `git`. They check:
  - A PR gets exactly one x64 job and every arm64 job.
  - The PR run and the main run together cover the full matrix, with no job running twice.
  - A push with no passing PR run of the same tree gets the full matrix.
  - After a full or matrix-skipped PR run, main runs nothing.
  - Each self-hosted job reads its matrix from the plan.
- Results: 28/28 pass, and `cargo fmt --check`, clippy and actionlint are clean.
- This PR's own CI run is the live check: it should show four arm64 jobs and one x64 job.

## Summary by CodeRabbit

- **CI Improvements**
  - Pull request validation now runs a focused test matrix while maintaining coverage across supported environments.
  - Main-branch builds run only the remaining checks when pull request coverage is available, reducing unnecessary duplicate work.
  - Builds run the complete matrix when prior pull request coverage is unavailable or incomplete.
  - Manual and workflow-triggered runs continue to validate the full matrix.

- **Tests**
  - Added coverage to verify matrix planning and execution behavior across supported event types.


ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's walkthrough for 28581a80a reports no actionable comments, minimal merge risk and 5 of 5 pre-merge checks passed; it carries no finding. Its one suggestion, "Fix failing CI checks", referred to Host-Root-x64-SnapshotEnabled in run 34783983989 attempt 1, whose runner (i-0cc85225bb0c40d3d, r5d.metal spot) was reclaimed by EC2 (Server.SpotInstanceTermination) mid test-root. The rerun in attempt 2 passed, and every check on this head is green.

PR #935, opened 2026-09-14
Drop the second vmm-sys-util from the lockfile

ejc3:
fuse-backend-rs pinned vmm-sys-util 0.12.1 while fcvm moved to 0.15.0, so every build compiled both. The fork now uses 0.15.0 (ejc3/fuse-backend-rs#1). This updates fuse-backend-rs to that commit, which removes 0.12.1 from Cargo.lock.

Tested: make build, make lint (cargo deny: advisories ok, bans ok, licenses ok, sources ok).


ejc3 (commented):
NOT-A-DEFECT: both top-level comments are notices that no review ran, not findings. The Codex comment reports its usage limit, and the CodeRabbit comment reports that it skipped the review because its path filters exclude Cargo.lock, the only file this PR changes.

PR #937, opened 2026-09-18
Bump rustls to 0.23.45 for RUSTSEC-2026-0285

ejc3:
`make lint` fails on every branch since the advisory was published on
2026-09-14: rustls 0.23.35 accepts TLS 1.3 handshake messages across
encryption level boundaries. reqwest pulls rustls in for fcvm and
fc-agent. The fix is in 0.23.45; rustls-webpki moves with it.

Tested:
  before: cargo audit       error: 1 vulnerability found!
          cargo deny check  advisories FAILED
  after:  cargo deny check  advisories ok, bans ok, licenses ok, sources ok
          cargo audit       no vulnerabilities (the allowed bincode
                            unmaintained notice remains)
          make clippy       Finished, no warnings


ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's only output on this PR is its notice that the review was skipped.

The one changed file, `Cargo.lock`, is excluded by CodeRabbit's default path filters, and Codex reported that its review quota is used up. Neither raised anything about the change.

The change moves rustls from 0.23.35 to 0.23.45 for RUSTSEC-2026-0285, with rustls-webpki following it. `cargo deny check` fails on main without it and passes with it; CI on this head is green.

PR #946, opened 2026-09-19
Health probes never ask a proxy for the guest

ejc3:
With `http_proxy` exported, a rootless or bridged VM started with `--health-check` never reads healthy. Two of the three HTTP health probes in `src/health.rs` take the proxy from fcvm's environment and ask it for the guest's address:

- The rootless probe runs `curl` inside the VM's namespace through `nsenter`, with no `--noproxy`. curl cannot reach the proxy from there and exits non-zero with nothing on stderr, so the monitor logs `Failed to connect to 10.0.2.100:80 via nsenter:` with nothing after the colon, for as long as the VM lives.
- The bridged probe is a `reqwest` client, and reqwest 0.12 reads `http_proxy` from the environment unless the builder says `no_proxy()`.

The routed probe already passes `--noproxy '*'` (09c1a168). A host that needs a proxy to pull images has these variables set for every fcvm process, so the only way to run a health-checked VM there is to list guest addresses in `no_proxy`.

## Evidence

One command on one host with `http_proxy` exported: nginx with `--publish` and `--health-check http://localhost:80`, rootless.

| `no_proxy` | result |
|---|---|
| without the guest address | still `unhealthy` after 7 minutes and 3,720 consecutive probe failures, with nginx serving the whole time |
| with `10.0.2.100` | healthy in 4.0 s |

## Change

`build_nsenter_curl_args` adds `--noproxy '*'`, and the bridged client is built with `.no_proxy()`. A guest address is never behind a proxy.

## Test results

Both tests were written first and run against the unfixed code. `nsenter_curl_never_asks_a_proxy_for_the_guest` checks the argument list. `bridged_probe_never_asks_a_proxy_for_the_guest` re-executes the test binary as a child with `http_proxy` pointing at a loopback listener, because this crate's tests do not change their own environment (`src/test_env.rs`). The parent plays a guest that answers 200 and a proxy that answers 502, and counts who was asked.

Red:

```
$ make test-unit FILTER="-E 'test(/never_asks_a_proxy/) | test(/no_test_mutates_the_process_environment/)'"
    no --noproxy in ["-t", "12345", "-U", "-n", "--preserve-credentials", "--", "curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", "--max-time", "5", "http://10.0.2.100:80/health"]
    assertion `left == right` failed: the health probe asked http_proxy for the guest (0 direct requests)
```

Green:

```
        PASS [   0.005s] (1/3) fcvm health::tests::nsenter_curl_never_asks_a_proxy_for_the_guest
        PASS [   0.011s] (2/3) fcvm test_env::tests::no_test_mutates_the_process_environment_outside_this_module
        PASS [   0.017s] (3/3) fcvm health::tests::bridged_probe_never_asks_a_proxy_for_the_guest
     Summary [   0.021s] 3 tests run: 3 passed, 1121 skipped
```

`make lint` passes.

## CI

At 746a54ad one job failed, `Host-Root-x64-SnapshotEnabled`, on `test_restore_network_cleanup_reestablishes_gateway` with `wget: can't connect to remote host (127.0.0.1): Connection refused`, on all four tries across two attempts. This change does not reach that path: CI exports no proxy and the test's fixture passes no health check. The same test failed its first try the same way on main's x64 root job at 6f06bd1d. The cause was the shared test fixture snapshotting nginx before it listens, fixed in #947. The head is now 746a54ad plus a merge of main (69b65f05), which brings that fix in.

## Summary by CodeRabbit

* **Bug Fixes**
  * Health checks now bypass configured HTTP proxies, ensuring requests reach the guest service directly.
  * Improved reliability for both bridged health checks and namespace-entered probes when proxy environment variables are present.


ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai review

ejc3:
@coderabbitai full review

ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's walkthrough for 69b65f05 reports no actionable comments

It covers 1c3c7fd4109da0644ce642b4fad88c83744391a1 and 69b65f05b458588e20ce5f4c9d5a8bb306c4f684 and raises no finding. The head is 746a54ad, which CodeRabbit already reviewed with no actionable comments, plus a merge of main that brings in the test fixture fix from #947. Its other replies are notices about its own state, not claims about the code.

PR #951, opened 2026-09-19
Parity test: replace the reference container when podman fails on its own twice

### LINKED ISSUE #944: test_exec_matches_podman_exec: the reference podman failed to find user nobody twice in a row
`test_exec_matches_podman_exec` failed its first try on main at 6f06bd1d (`Host-Root-x64-SnapshotEnabled`, run 35409538443) and passed on nextest's retry. The failure is on the reference side: the runner's own `podman exec -u nobody` against the reference container failed, twice in a row.

```
user_by_name   podman reported [70 bytes] "Error: unable to find user nobody: no matching entries in passwd file\n"; asking again
...
66 cases, 6 differences
user_by_name [fcvm exec]      stdout: podman [0 bytes] ""  fcvm [18 bytes] "65534\n65534\n65534\n"
user_by_name [fcvm exec]      stderr: podman [70 bytes] "Error: unable to find user nobody: no matching entries in passwd file\n"  fcvm [0 bytes] ""
user_by_name [fcvm exec]      exit:   podman Some(255)  fcvm Some(0)
(the same three for fcvm exec --vm)
```

fcvm's answer is the right one: alpine has `nobody` as 65534. This is not the guest scheduler stall that #940 fixed. No exec in that try was slow and nothing timed out.

The test already knows about this. It asks podman again when podman's own stderr starts with `Error:`, with the comment "podman's own failures are sometimes transient under load (seen: a user lookup in the image failing once). One that is real repeats." Here it repeated on the second ask, within the same reference container, and was gone in the retry, which starts a fresh reference container. So one retry of the case is not enough, and "transient under load" is not yet an explanation.

## What is known

- Every other case in that try matched, including `user_and_group_by_number` and `user_unknown_number`, which need no passwd lookup.
- The same job's log has `attempt to write a readonly database` from podman inside several guests during that window. Those are guest-side and a different podman, so they are not evidence about the host's.
- The only host-side podman command in the job that touches storage is one `podman system migrate || true` during setup.

## What would attribute it

When the reference fails a user lookup, capture at that moment, from the host: `podman exec <ref> cat /etc/passwd`, `podman inspect --format '{{.State.Status}} {{.GraphDriver.Data.MergedDir}}' <ref>`, and whether that merged directory still has `etc/passwd`. That separates a reference container whose root is briefly unmounted or replaced from a podman lookup bug. Until it is attributed, the honest options are to make the reference independent of the lookup for this case (compare against a fixed expectation) or to restart the reference container and rerun the case, and to say which in the test.

ejc3:
`test_exec_matches_podman_exec` no longer fails when host podman fails on its own twice in one reference container. It collects evidence from the host, replaces the reference container, and asks the case once more.

## The defect

The test compares `fcvm exec` with host `podman exec`. When podman answered a case with its own `Error:`, the test asked once more and compared fcvm with whatever the second ask said. In #944 `podman exec -u nobody` answered `Error: unable to find user nobody: no matching entries in passwd file` twice in a row, inside one reference container, for an image that has the user. fcvm printed `65534`, which is right, and the test reported six differences that were podman's. nextest's retry starts a fresh reference container, and it passed.

It is not rare. Of six root-job logs from runs 35424243337, 35424262960 and 35433024229, five show this exact first-try failure (`66 cases, 6 differences`, the reference exiting 255 for `user_by_name` on both asks) and a passing second try, on arm64 and on x64. None of them says why: the failure carried no evidence.

Four cases make podman print `Error:` on every run, by design: `plain_command_not_found`, `plain_command_not_executable`, `workdir_missing` and `i_command_not_found`. fcvm fails the same way, so for them the repeated failure is the answer. Any fix has to leave them alone.

## The change

Test code only. The decision moves to `tests/parity_reference/mod.rs`, a function over closures (ask, is this podman's own failure, compare, collect evidence, restart, note), so it can be tested with no podman and no VM. `tests/test_parity_reference.rs` holds those tests and carries no feature gate, so `make test-unit` runs them.

1. An own failure is asked for again, as before. Evidence is collected from the host before the second ask, because a failure that does not repeat leaves nothing to look at later:
   - `podman exec <ref> cat /etc/passwd`: exit code, first lines, and whether there is a `nobody` entry. `cat` runs as the image's default user, so no lookup is involved.
   - `podman inspect --format '{{.State.Status}} {{.GraphDriver.Data.MergedDir}}' <ref>`.
   - Whether `<MergedDir>/etc/passwd` exists. As root that is a `stat` from the host. Rootless it goes through `podman unshare`, because rootless podman mounts the merged root only in its own namespace and the host's view of that path is empty whatever state the container is in. I measured both on podman 5.8.5.
   - Every probe has a 20 s limit and writes to temporary files. One that cannot start, times out, or ends without an exit code says so in the output.
2. A failure that is gone on the second ask prints that evidence and the case goes on.
3. A failure that repeats and that fcvm shares is the case's answer. Nothing else happens and nothing is printed, so the four cases above stay quiet.
4. A failure that repeats while fcvm disagrees gets a second evidence block, a fresh reference container, and one more ask. This is the "restart the reference and rerun the case" option from #944. The other option, a fixed expectation for the case, would stop comparing it with real podman, and that comparison is the point of the test. The new container is up before the old one is removed, and a start that fails is an error for the case, not a comparison with a stale answer.
5. Only if the fresh container fails the same way do the differences stand. The evidence is attached to the failure message, after the differences count, the way the "stopped at the limit" notice is.

The test also runs the collector once against the healthy reference container and requires it to see the `nobody` entry, the `running` state and the merged file, and then replaces the reference once. A collector that cannot see a healthy container would say nothing useful about a failing one, and neither path should first run on the day it is needed.

## Red and green

The unit tests against the decision as it is on main (ask, ask again, compare), behind the same signature:

```
$ make test-unit FILTER="-E 'binary(test_parity_reference)'"
    Starting 7 tests across 1 binary (97 binaries skipped)
        PASS  an_answer_that_is_not_a_failure_is_asked_for_once
  TRY 1 FAIL  a_failure_a_fresh_reference_does_not_repeat_is_not_a_difference
  TRY 1 FAIL  a_failure_a_fresh_reference_repeats_fails_and_carries_the_evidence
  TRY 1 FAIL  a_failure_that_is_gone_on_the_second_ask_shows_its_evidence_and_needs_no_restart
  TRY 1 FAIL  a_failure_the_target_shares_is_the_answer_and_needs_no_restart
  TRY 1 FAIL  a_restart_that_fails_stops_the_case_after_the_evidence_is_out
  TRY 1 FAIL  a_target_difference_survives_a_fresh_reference
     Summary 7 tests run: 1 passed, 6 failed, 0 skipped
```

The first of those is #944 itself, and on main's logic it reports the difference the issue shows:

```
assertion `left == right` failed
  left: ["stdout: podman \"Error: unable to find user nobody: no matching entries in passwd file\\n\" fcvm \"65534\\n65534\\n65534\\n\""]
 right: []
```

With the change:

```
     Summary [   0.008s] 7 tests run: 7 passed, 0 skipped
```

Two of the tests guard against doing too much, so main's logic cannot fail them for the right reason. Each was watched failing under a mutation of the change, and the change was restored afterwards (`cmp` against the saved copy):

| mutation | tests that fail |
|---|---|
| replace the container whenever the failure repeats, even when fcvm agrees | `a_failure_the_target_shares_is_the_answer_and_needs_no_restart` |
| print the evidence for a failure fcvm shares | the same test, on its "stays quiet" assertion |
| skip the second ask | six of seven, including `a_failure_that_is_gone_on_the_second_ask_...` |
| always ask twice | all seven, including `an_answer_that_is_not_a_failure_is_asked_for_once` |

`make lint` is clean. clippy does compile this feature-gated test: a deliberate unused variable inside `test_exec_matches_podman_exec` failed `make clippy` with `could not compile fcvm (test "test_exec_podman_parity")`, and was removed.

The real test, once, as root on an x86_64 host with podman 5.8.5:

```
$ make _test-root FILTER=test_exec_matches_podman_exec
    the reference container, healthy:
    $ podman exec exec-parity-clone-2478083-0 cat /etc/passwd
      exit 0
      stdout | root:x:0:0:root:/root:/bin/sh
      stdout | bin:x:1:1:bin:/bin:/sbin/nologin
      stdout | daemon:x:2:2:daemon:/sbin:/sbin/nologin
      stdout | (17 lines, first 3 shown)
      a nobody entry: yes
    $ podman inspect --format {{.State.Status}} {{.GraphDriver.Data.MergedDir}} exec-parity-clone-2478083-0
      exit 0
      stdout | running /var/lib/containers/storage/overlay/9a2e8272c6b2.../merged
    /var/lib/containers/storage/overlay/9a2e8272c6b2.../merged/etc/passwd: present, 702 bytes, seen from the host

    the reference container is now exec-parity-clone-2478083-0-r1
    ...
    66 cases, 0 differences
        PASS [ 113.600s] (1/1) fcvm::test_exec_podman_parity test_exec_matches_podman_exec
```

The four cases whose failure fcvm shares asked twice, printed no evidence and replaced nothing. The lookup did not fail in that run, so the recovery itself has only run against the scripted reference. Five of the last six root jobs hit the failure on their first try, so this PR's own root jobs are likely to show it: look for `again and the target disagrees` in their logs. That output is also what attributes the podman side, which is still unknown. If it shows the merged root gone from the host, that is a host-side podman problem and deserves its own issue.

Closes #944


ejc3:
Closing in favour of #959. The cause of #944 turned out to be in fcvm: the temporary-store `podman load` in `build_storage_image` shared the default runroot, which detaches every running container of the default store, the parity test's reference container among them. Restarting the reference and letting the case pass, as this PR does, would have hidden that defect. #959 fixes the cause and pins it with a test that fails without the fix. A smaller change that only attaches evidence to a parity failure, with no restart and no change of verdict, follows separately.

PR #963, opened 2026-09-20
uffd tests: end the stand-in VMM on every exit path, a panic included

ejc3:
## The defect

Five tests in `src/uffd/server.rs` spawn a `sleep 600` to stand in for a clone's VMM (`spawn_victim`) and end it only at their last step. `std::process::Child` does nothing on drop, so a panic anywhere between the spawn and that step leaves the `sleep` running for up to ten minutes.

`test_handler_failure_kills_the_clones_vmm` hits this wherever `/dev/userfaultfd` is not accessible to the test user. It spawns the victim, then `UffdBuilder::create()` fails with `Permission denied` and the `expect` panics. nextest reports the test as `FL+LK`, a failure that also leaked handles. The orphan keeps every descriptor it inherited from the test runner, including the shared lock the cargo wrapper holds on the checkout's target directory, and `scripts/cargo-target-link.sh` takes that lock exclusively at the start of every `make`. Measured on the development host: a `make` issued after an unprivileged unit run waited in `flock -x` for about 6 minutes, until the `sleep` exited.

Where the device is accessible the test passes and ends its victim. CI is such a place: its workflow runs `sudo chmod 666 /dev/userfaultfd` on the runners.

## The fix

`spawn_victim` returns a `Victim`, a newtype over `Child` that derefs to it, so no call site changes. Two things end it, because neither covers every exit alone:

- **Drop** kills the process and reaps it. That covers a test that returns and a test that panics, since unwinding drops it.
- **`PR_SET_PDEATHSIG`**, armed in the child by a `pre_exec` hook with the same fork-versus-prctl check `tests/common` uses for every fcvm it spawns. That covers a test process that is killed outright, which is what nextest does to a test that outlives its timeout: nothing unwinds, so no drop runs. AGENTS.md ("Process teardown is per-hop") asks for the kernel hop wherever teardown has to survive SIGKILL.

The guard acts only when it goes out of scope. `assert_vmm_was_killed` takes the victim by value and reads how it died before letting go of it, so the guard's own SIGKILL cannot stand in for the one a test asserts on.

All five callers are covered by the type: `test_peer_vmm_kill_targets_only_the_pinned_process`, `admitted_task_panic_kills_the_pinned_vmm`, `dropping_unpolled_admitted_task_kills_the_pinned_vmm`, `test_handler_failure_kills_the_clones_vmm` and `test_read_event_error_kills_the_clones_vmm`. `spawn_victim` is the only place the file's tests spawn a process.

## Red, then green

Two tests, one per mechanism. Both use `spawn_victim` only through what it offers on main, so the identical test code runs on the unfixed tree. The oracle is a pidfd (`PeerVmm::from_pid`), which names the exact process.

- `a_victim_does_not_outlive_a_test_that_panics`. A thread spawns a victim and panics, as a failing test body does. The victim must be gone and reaped: a pidfd still reads alive for a zombie, so a guard that killed without waiting fails this too.
- `a_victim_whose_guard_never_runs_dies_with_the_thread_that_spawned_it`. The thread forgets the victim and returns, which is what a killed test process amounts to for a drop guard. The pidfd must poll readable within 5 s, and the wait status must be SIGKILL.

Both end a stray victim before they assert, so a red run does not leak the process they exist to catch.

Red, on main with only the two tests added. The run is unprivileged, on a host whose `/dev/userfaultfd` is `crw------- root root`, so the nine tests already in `src/uffd` that create a userfaultfd fail as well, each with `Permission denied`. One of those nine is the test that leaks, and nextest marks it `FL+LK`:

```
$ make test-unit FILTER="--no-fail-fast -E 'test(/uffd::/)'"
  TRY 2 FAIL [   0.006s] (69/81) fcvm uffd::server::tests::a_victim_does_not_outlive_a_test_that_panics
 TRY 2 FL+LK [   0.209s] (80/81) fcvm uffd::server::tests::test_handler_failure_kills_the_clones_vmm
  TRY 2 FAIL [   5.012s] (81/81) fcvm uffd::server::tests::a_victim_whose_guard_never_runs_dies_with_the_thread_that_spawned_it
     Summary [  15.032s] 81 tests run: 70 passed, 11 failed, 1063 skipped

thread 'uffd::server::tests::a_victim_does_not_outlive_a_test_that_panics' panicked at src/uffd/server.rs:3568:9:
the stand-in VMM (pid 2638269) was still there after the test that spawned it panicked, and nothing else would have ended it for the rest of its 600 s

thread 'uffd::server::tests::a_victim_whose_guard_never_runs_dies_with_the_thread_that_spawned_it' panicked at src/uffd/server.rs:3613:9:
the stand-in VMM (pid 2642897) was still running 5 s after the thread that spawned it returned without dropping it; a test process that is killed outright would leave it behind for the rest of its 600 s

thread 'uffd::server::tests::test_handler_failure_kills_the_clones_vmm' panicked at src/uffd/server.rs:3776:14:
creating userfaultfd (via /dev/userfaultfd): OpenDevUserfaultfd(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })
```

The 11 failures are those nine and the two new tests. The test code is byte-identical in the red and green runs.

The orphans themselves, counted after each of those runs: `sleep 600` processes of the test user that were reparented to init, appeared since a snapshot taken just before the run, and hold a descriptor on this checkout's cargo target directory. The last condition matters on a shared host: the second run's snapshot also caught two `sleep 600` orphans whose working directory and target-directory descriptor belonged to a different checkout, and they are not counted.

```
without the fix:
  orphan: 2635573  1  15 sleep 600   descriptors on the target directory: 1
  orphan: 2638327  1  10 sleep 600   descriptors on the target directory: 1
with the fix:
  none
```

There are two because nextest retries a failed test once. The columns are pid, parent pid, age in seconds and command line.

Green, with the fix, same command and host. The nine tests that need the device still fail and nothing else does. `test_handler_failure_kills_the_clones_vmm` is now a plain `FAIL` with no leak:

```
$ make test-unit FILTER="--no-fail-fast -E 'test(/uffd::/)'"
        PASS [   0.007s] ( 6/81) fcvm uffd::server::tests::a_victim_does_not_outlive_a_test_that_panics
        PASS [   0.010s] (41/81) fcvm uffd::server::tests::a_victim_whose_guard_never_runs_dies_with_the_thread_that_spawned_it
  TRY 2 FAIL [   0.007s] (73/81) fcvm uffd::server::tests::test_handler_failure_kills_the_clones_vmm
     Summary [   5.051s] 81 tests run: 72 passed, 9 failed, 1063 skipped
```

As root, where the device opens, the whole module passes, including all five tests that spawn a victim:

```
$ make _test-root FILTER="--no-fail-fast -E 'test(/uffd::/)'"
     Summary [   0.149s] 81 tests run: 81 passed, 1387 skipped
$ make lint
```

## Same shape elsewhere, not touched here

Other tests spawn a long `sleep` as a plain `Child` and end it at their last step, after calls that can panic. I read these and did not change them, and I have not seen any of them leak.

- `tests/test_cgroup_swap.rs`: both tests (`sleep 300`, three children in all) read cgroup files with `expect` and assert on them before they kill and reap.
- `src/network/pasta.rs`, `live_child()`: `sleep 3600` as a tokio `Child` with no `kill_on_drop`. Three of its four callers end it with `child.kill().await` as their last step, after their assertions. The fourth, `pid_file_timeout_waits_for_complete_stderr`, never kills it and relies on the timeout branch under test to do so.
- `tests/test_fuzz_chaos.rs`, `a_dropped_clone_guard_kills_the_clone_it_guards`: `sleep 300`, tokio, no `kill_on_drop`. If the guard under test does not kill the child, the `expect` that reports it panics and nothing else ends the child.
- `tests/test_signal_cleanup.rs`, `test_zombie_is_not_running_regression_628`: `sleep 1000`, with two fallible steps between the spawn and its SIGKILL.

Two are already covered. `src/hypervisor/cloud_hypervisor/mod.rs` sets `kill_on_drop(true)`. `working_set_replay_cleans_isolation_clones_after_failed_assertion` in `tests/test_snapshot_clone.rs` kills its stand-in through a pidfd and reaps it before it asserts, which is the pattern the two new tests here follow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of test processes to prevent orphaned background processes.
  * Added safeguards to terminate spawned processes when their owner exits unexpectedly.
* **Tests**
  * Added coverage for process cleanup after test failures.
  * Added coverage for cleanup when the spawning thread exits without explicitly releasing the process.


ejc3 (commented):
NOT-A-DEFECT: neither bot review of 2d78e761 carries a finding.

Codex completed its review of this head and left no comments. CodeRabbit's walkthrough for the same head reports no actionable comments.

PR #965, opened 2026-09-20
Tests: helper processes share one guard and cannot outlive a failed test

ejc3:
## The defect

A test that needs a live process to aim code at spawns `sleep`. Neither `std::process::Child` nor `tokio::process::Child` ends its process when dropped, so a test that fails between the spawn and its own cleanup leaves the `sleep` running. The orphan keeps every descriptor it inherited, including the lock `make` holds on the worktree's cargo target directory, so the next `make` in that checkout blocks in `flock -x` until the sleep runs out. Measured while working on #963: a `make` issued after an unprivileged unit run waited in `flock -x` for about 6 minutes.

#963 fixed this for the five uffd tests. Its review found the same leak in these places:

- `src/network/pasta.rs`, `live_child()`: tokio `sleep 3600` with no `kill_on_drop` and no pdeathsig. Its four callers have expects and asserts between the spawn and `child.kill().await`, so a failure there leaves a one-hour orphan.
- `tests/test_fuzz_chaos.rs`, `a_dropped_clone_guard_kills_the_clone_it_guards`: `sleep 300`, leaked exactly on the path where the test fails (`expect("guard did not kill the clone ...")`).
- `tests/test_signal_cleanup.rs`, `test_zombie_is_not_running_regression_628`: `sleep 1000`, with an `expect` and an `assert!` before the kill.
- `tests/test_cgroup_swap.rs`: three `sleep 300` in the privileged lane, each with expects and asserts before the cleanup at the end.
- `tests/test_snapshot_clone.rs`, `working_set_replay_cleans_isolation_clones_after_failed_assertion`: a `sleep 600` stand-in with an `assert!` between the spawn and the hand-off. This one was not in the review. A sweep of every `Command::new("sleep")` in the workspace found it.

## The fix

**One guard, one implementation.** `src/test_child.rs` holds `TestChild<C>`, a guard over a std or a tokio child that dereferences to it. `spawn_sleep(seconds)` returns the tokio flavour and `spawn_sleep_std(seconds)` the std one. Both spawn `sleep` with no stdio and arm `PR_SET_PDEATHSIG` in a `pre_exec` hook that re-checks `getppid`. Dropping the guard kills the process and reaps it.

The guard reaps by PID, which is safe because the child is the test's own and unreaped: `Child::wait` for a std child, and for a tokio child, which reaps only from an async `wait` that a drop cannot await, `start_kill`, then `waitid(P_PID, WEXITED | WNOWAIT)` until the kernel reports the exit, then `try_wait` so tokio collects the status. If the kill fails nothing was signalled, so the drop says so and returns instead of waiting out the sleep. Only `Pinned`, the handle the tests of the guard use to ask whether a process is gone, goes through a pidfd (`pidfd_open`, `pidfd_send_signal`, `waitid(P_PIDFD)`).

The library declares the file as a `cfg(test)` module. Integration tests link the library without its `cfg(test)` modules, so `tests/common` includes the same file with `#[path = "../../src/test_child.rs"] pub mod test_child;`. The file refers to nothing else in the crate, and it compiles in every integration test binary under the three feature sets the lanes use. There is no second copy.

- `src/uffd/server.rs`: `Victim` is an alias for `TestChild<std::process::Child>` and `spawn_victim()` calls `spawn_sleep_std(600)`. The inline guard and hook from #963 are deleted.
- `connect_stand_in_vmm`, the warm-up tests' stand-in from #960 that connects to the server's socket between fork and exec, starts from the shared `sleep_command(600)` and adds its connect step as a second `pre_exec` hook. Hooks run in the order they were added, so the child is tied to the test thread before it starts connecting. It guards the child with `Victim::new`, and its own `die_with_the_test` is deleted.
- `src/network/pasta.rs`: `live_child()` returns `TestChild<Child>`. The callers still lend `&mut child` to `wait_for_pid_file`, through the deref. The three `child.kill().await.expect("stop live child")` lines are gone because the guard does that on every path.
- `tests/`: the sleeps in `test_fuzz_chaos.rs` (both clone guard tests), `test_signal_cleanup.rs` and `test_cgroup_swap.rs` use `common::test_child`.
- Two children move by value into the code under test, where a guard that kills on drop would hide the very leak the test looks for: the stand-in in `test_snapshot_clone.rs` and the `sleep 60` that `reboot_monitor_failure_reaps_child_before_returning` puts into the Cloud Hypervisor backend. They get the kernel half alone, `die_with_the_spawning_thread(command)`.

**The pdeath signal is SIGTERM for every sleep.** `assert_vmm_was_killed` in the uffd tests and `a_dropped_clone_guard_kills_the_clone_it_guards` both recognise the kill under test by SIGKILL. With SIGKILL as the pdeath signal too, a child spawned by mistake from a short-lived helper thread would die by signal 9 when that thread returned and satisfy the assertion without that kill. `set_test_pdeathsig` and `set_test_pdeathsig_std` are unchanged. They arm the fcvm processes that integration tests spawn, with SIGKILL.

**The other follow-ups from the review of #963**, in the uffd test code:

1. The two guard tests reaped with `waitpid(peer.pid)`, a field `PeerVmm` documents as for logs only. They now hold a `Pinned` and reap through its pidfd.
2. The helper thread that panics on purpose is named `panics-on-purpose` through `thread::Builder`. nextest prints passing output immediately here, so every green run printed `thread '<unnamed>' panicked`.
3. The timeout branch of `assert_vmm_was_killed` no longer kills and waits by hand. The panic drops the victim and the guard does both.

No production code path changes. `tests/test_serve.rs` gates its `Duration` import on `integration-slow` like the one test that uses it, which removes the unused import warning the unit lane printed.

## Red, then green

Written first, against the unfixed `live_child()`: a named thread obtains the live child inside a runtime of its own and panics before ending it, the way a failing test body does. After the join the child must be gone and reaped. A pidfd still reads present for a zombie, so a guard that kills without reaping fails this too. A second test forgets the child instead, which is what a killed test process amounts to for a drop guard, and requires the kernel to end it with SIGTERM within 5 s. Both end a stray child before they assert, so the red run leaked nothing.

```
$ make test-unit FILTER="-E 'test(/a_live_child/)'"
  TRY 1 FAIL [   0.007s] fcvm network::pasta::tests::a_live_child_does_not_outlive_a_test_that_panics
thread 'panics-on-purpose' panicked at src/network/pasta.rs:2735:13:
stand-in for an expect that fails before the child is ended
thread 'network::pasta::tests::a_live_child_does_not_outlive_a_test_that_panics' panicked at src/network/pasta.rs:2749:9:
the live child (pid 253803) was still there after the test that spawned it panicked, and nothing else would have ended it for the rest of its 3600 s
  TRY 1 FAIL [   5.010s] fcvm network::pasta::tests::a_live_child_whose_guard_never_runs_dies_with_the_thread_that_spawned_it
thread 'network::pasta::tests::a_live_child_whose_guard_never_runs_dies_with_the_thread_that_spawned_it' panicked at src/network/pasta.rs:2783:9:
the live child (pid 253804) was still running 5 s after the thread that spawned it returned without dropping it; a test process that is killed outright would leave it behind for the rest of its 3600 s
     Summary [  15.025s] 2 tests run: 0 passed, 2 failed, 1148 skipped
```

The test code is byte-identical in the red and green runs. Only `live_child()` changed.

The pair in `tests/test_signal_cleanup.rs` reaches the same file through the include. It was watched failing too, with the shared guard's drop emptied and the `prctl` removed from its hook:

```
$ make _test-fast FILTER="-E 'test(/test_spawned_sleep/)'"
  TRY 1 FAIL [   0.009s] fcvm::test_signal_cleanup test_spawned_sleep_does_not_outlive_a_test_that_panics
the std sleep (pid 1489581) was still there after the test body that spawned it panicked, and nothing else would have ended it for the rest of its 1000 s
  TRY 1 FAIL [   5.014s] fcvm::test_signal_cleanup test_spawned_sleep_whose_guard_never_runs_dies_with_its_spawning_thread
the std sleep (pid 1489569) was still running 5 s after the thread that spawned it returned without dropping it; a test process that is killed outright would leave it behind for the rest of its 1000 s
     Summary [  15.033s] 2 tests run: 0 passed, 2 failed, 1248 skipped
```

With the file restored (its checksum matched the one taken before the mutation) both pass.

## Tests

New or changed:

- `network::pasta::tests::a_live_child_does_not_outlive_a_test_that_panics` (new)
- `network::pasta::tests::a_live_child_whose_guard_never_runs_dies_with_the_thread_that_spawned_it` (new)
- `uffd::server::tests::a_victim_does_not_outlive_a_test_that_panics` (named thread, reaps through the `Pinned` pidfd)
- `uffd::server::tests::a_victim_whose_guard_never_runs_dies_with_the_thread_that_spawned_it` (asserts SIGTERM, reaps through the `Pinned` pidfd)
- `test_signal_cleanup::test_spawned_sleep_does_not_outlive_a_test_that_panics` (new, the std and the tokio flavour through `tests/common`)
- `test_signal_cleanup::test_spawned_sleep_whose_guard_never_runs_dies_with_its_spawning_thread` (new, same two flavours, asserts SIGTERM)

Their sleeps changed and their assertions did not: `hypervisor::cloud_hypervisor::tests::reboot_monitor_failure_reaps_child_before_returning`, `test_signal_cleanup::test_zombie_is_not_running_regression_628`, `test_fuzz_chaos::a_dropped_clone_guard_kills_the_clone_it_guards`, `test_fuzz_chaos::a_disarmed_clone_guard_leaves_the_process_alone`, both `test_cgroup_swap` tests and `test_snapshot_clone::working_set_replay_cleans_isolation_clones_after_failed_assertion`.

Runs, on this branch rebased on 852e5530 (main with #952, #960 and #962):

- `make lint`: clean. clippy builds every integration test binary with default features, the included file among them, with no warning.
- `make test-unit` as an unprivileged user: 1206 run, 1191 passed, 14 failed, 1 timed out. The 14 failures are the `uffd::` tests that create a userfaultfd, each with `Permission denied`, on a host whose `/dev/userfaultfd` is `crw------- root root`: the nine #963 listed and five that #962 added. The timeout is `test_probe_pasta_dns_gateway::the_probe_leaves_no_dns_responder_behind`. It times out the same way in unprivileged runs of other branches on that host, and passes in CI. The three tests that use `connect_stand_in_vmm` pass (`a_dropped_stand_in_vmm_is_killed_and_reaped`, `admitting_a_clone_warms_the_recorded_set_again`, `the_stand_in_vmm_waits_out_a_socket_that_is_bound_but_not_yet_listening`). The build printed no compiler warning. Before, it printed `unused import: std::time::Duration` for `tests/test_serve.rs`.
- After that run, `ps -eo pid=,ppid=,etimes=,args= | awk '$2==1 && $4=="sleep"'` printed nothing. The only sleeps of 5, 60, 300, 600, 1000 or 3600 seconds younger than the run were children of unrelated shell scripts that were still running as their parents. Before #963 the same run left `sleep 600` orphans behind.
- The privileged lane for every test touched here, built with `--features privileged-tests` the way `make test-root` does: `make _test-root FILTER="-E 'test(/pasta::tests|uffd::server::tests|reboot_monitor_failure_reaps_child|test_disable_cgroup_swap|test_spawned_sleep|test_zombie_is_not_running|clone_guard|cleans_isolation_clones/)'"`: 120 run, 120 passed. That includes the `uffd::server` tests that need the device (`test_handler_failure_kills_the_clones_vmm` among them), the three `connect_stand_in_vmm` tests, both cgroup tests, both clone guard tests, the Cloud Hypervisor test and the snapshot clone stand-in test. Its build prints 10 dead code warnings, all in `tests/test_kvm.rs`, which this change does not touch.
- The unprivileged fast lane for the `tests/` tests touched here: `make _test-fast FILTER="-E 'test(/test_spawned_sleep|test_zombie_is_not_running|clone_guard/)'"`: 5 run, 5 passed, no compiler warning, and no `sleep` left afterwards.

## What remains of the same shape

- `tests/test_egress_stress.rs` (`start_http_server`) and `tests/test_rootless_ipv6.rs` spawn a `python3` HTTP server as a bare tokio child and kill it only at the end of the test. A failure before that leaves a server that never exits on its own. Both are VM tests that could not run where this was written, so they are not touched here.

## Summary by CodeRabbit

* **Tests**
  * Improved cleanup of helper processes spawned during tests, including when tests panic or their spawning threads exit unexpectedly.
  * Added coverage verifying that test-created processes do not outlive failed or interrupted test runs.
  * Simplified process cleanup across several test suites by using shared lifecycle management.
  * Reduced test warnings when optional integration features are disabled.


ejc3:
@codex review

ejc3 (commented):
NOT-A-DEFECT: the review of bc6bf9c3 above reports no findings, so there is nothing to fix and nothing to verify.

ejc3:
@codex review

ejc3 (commented):
NOT-A-DEFECT: the review of 222f9283 above reports no findings, so there is nothing to fix and nothing to verify.

ejc3:
@codex review

ejc3 (commented):
NOT-A-DEFECT: the review of abc39bf1 above reports no findings, so there is nothing to fix and nothing to verify.

ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's summary comment, as edited at 2026-09-20T09:14:22Z for abc39bf1, says no actionable comments were generated, so there is nothing to fix and nothing to verify.

Issue #969, opened 2026-09-20
Firecracker runs at Debug, so a vsock-heavy guest writes gigabytes an hour into firecracker.log


ejc3:
fcvm starts Firecracker with `--level Debug` (`src/firecracker/vm.rs`, next to `--log-path`). At that level Firecracker's vsock muxer writes one line per packet, so a guest whose volumes are FUSE over vsock fills the per-VM `firecracker.log` with packet headers.

## Measured

A 128 GiB guest with FUSE-over-vsock volumes, still warming up. 65 minutes after boot:

```
size 5.10 GiB
3712 lines/s over the newest 2 MiB, every one of them
  [anonymous-instance:main:DEBUG:src/vmm/src/devices/virtio/vsock/unix/muxer.rs:205] vsock: muxer.send[rxq.len=0]: VsockPacketHeader { src_cid: 3, dst_cid: 2, ... }
```

Nineteen minutes later the file was 9.99 GiB, so the rate follows the guest's vsock traffic: about 78 MiB a minute on average over the first hour, and about 260 MiB a minute in those nineteen. A `podman prepare` of that guest takes about 104 minutes to turn ready, and its log was 7 to 8 GiB by then, on the filesystem the snapshot is about to be written to (#968 is a prepare that ran out of room there). A VM that stays up grows the file without bound.

At teardown `cleanup_vm` copies the whole file to `/tmp/fcvm-firecracker-<vm_id>.log` before it removes the VM's directory, so every VM exit leaves a copy of that size in `/tmp`, and nothing removes those. One test host's `/tmp` held 44 of them, 23 GiB together, after two days, and a development host that runs the test suite held 2,820 of them.

The cost that is established is disk. A FUSE workload in a small VM (a 2 GiB sequential read, 20,000 files of 1 KiB, a 1 GiB write, through one `--map` volume), one fcvm binary with the level taken from an environment variable, three alternating pairs on a shared host:

| | Debug | Info |
|---|---|---|
| `firecracker.log` per run | 311, 311, 312 MiB | 0 MiB |
| 2 GiB read | 18.3, 22.9, 21.0 s | 21.6, 17.4, 16.4 s |
| 20,000 small files | 38.6, 36.1, 39.1 s | 38.3, 36.3, 33.3 s |
| 1 GiB write | 18.0, 16.6, 19.7 s | 17.0, 16.5, 15.9 s |

[...]

PR #970, opened 2026-09-20
Firecracker logs at Info unless FCVM_FIRECRACKER_LOG_LEVEL asks for more

ejc3:
## Firecracker logs at Info unless FCVM_FIRECRACKER_LOG_LEVEL asks for more

fcvm started every Firecracker with `--level Debug`. At that level Firecracker's vsock muxer writes one line per packet, so a guest whose volumes are FUSE over vsock fills the per-VM `firecracker.log` with packet headers. #969 has the measurements: a 128 GiB guest wrote 5.10 GiB in its first 65 minutes and 9.99 GiB by minute 84, onto the filesystem its snapshot was about to need, and teardown copies the whole file to `/tmp`.

## The change

`VmManager::start` passes `--level Info` by default. `FCVM_FIRECRACKER_LOG_LEVEL` names another level (Error, Warning, Info or Debug, which are the values Firecracker lists, in any case). A value that is none of those stops the command right after argument parsing, before it prepares a disk or a namespace, and `VmManager::start` checks again for callers of the library. The error names the variable and the accepted values. The README's environment table lists the variable. `Info` keeps Firecracker's API requests, warnings and errors.

Nothing in fcvm, its tests, its scripts or its workflows reads the Debug lines. The Cloud Hypervisor backend is unchanged.

Not in this PR, left in #969: bounding what teardown copies to `/tmp`, and a size limit for the log of a long-lived VM.

## Measured

One fcvm binary with the level taken from the environment, a FUSE workload in a small VM through one `--map` volume (2 GiB sequential read, 20,000 files of 1 KiB, 1 GiB write), three alternating pairs on a shared host:

| | Debug | Info |
|---|---|---|
| `firecracker.log` per run | 311, 311, 312 MiB | 0 MiB |
| 2 GiB read | 18.3, 22.9, 21.0 s | 21.6, 17.4, 16.4 s |
| 20,000 small files | 38.6, 36.1, 39.1 s | 38.3, 36.3, 33.3 s |
| 1 GiB write | 18.0, 16.6, 19.7 s | 17.0, 16.5, 15.9 s |

The means are 5 to 11 percent lower at `Info` and the ranges overlap, so this PR claims the disk and nothing about speed.

## Tests

Both new tests were watched failing while the default was still `Debug`:

```
firecracker::vm::tests::firecracker_logs_at_info_unless_asked_otherwise
  left: "Debug"   right: "Info"
firecracker::vm::tests::start_passes_the_log_level_to_firecracker
  ["--api-sock", ".../api.sock", "--log-path", ".../firecracker.log", "--level", "Debug", "--show-level", "--show-log-origin"]
  left: Some("Debug")   right: Some("Info")
```

The second test starts a fake Firecracker that records its arguments: the default is `Info`, `debug` from the environment becomes `Debug`, and `verbose` fails `start` with the fake never spawned.

A third test runs the binary: with `FCVM_FIRECRACKER_LOG_LEVEL=verbose`, `fcvm ls` exits non-zero and prints `FCVM_FIRECRACKER_LOG_LEVEL="verbose" is not a Firecracker log level; use one of Error, Warning, Info, Debug`. It was watched failing before `main` called the check: `fcvm ls` exited 0. That check came from review, which found that `start` runs only after a cold `podman run` has prepared its disk and spawned its namespace holder.

With the default at `Info`, on this branch's tree:

```
make lint                      exit 0 (advisories ok, bans ok, licenses ok, sources ok)
the three new tests            3 run, 3 passed (with the rest of test_cli_parsing: 9 run, 9 passed)
make test-unit                 1249 run, 1235 passed, 14 failed
make test-root FILTER=sanity   6 run, 4 passed
```

All 14 unit failures are `uffd::` tests that cannot open `/dev/userfaultfd` where it is root-only, and the same 14 fail on main on that host. `test_sanity_rootless` and `test_sanity_chrony_ntp_configured` booted VMs at the new default, and the two Firecracker logs they saved hold 37 and 59 lines with no DEBUG line. `test_sanity_bridged` and `test_sanity_routed` fail on that host before Firecracker is started ("this host has no IPv4 default route", "routed networking requires a host with a global IPv6 address"), and fail the same way on main (c7434307). CI runs them on hosts that have both.


ejc3:
@codex review

ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's summary comment for a220b3ae says no actionable comments were generated, and Codex's review of a220b3ae reports no major issues, so there is nothing to fix and nothing to verify.

PR #977, opened 2026-09-21
fc-agent: --forward-localhost also listens on ::1

ejc3:
The guest relay for `--forward-localhost` bound `127.0.0.1` only, so a client that resolves `localhost` to `::1` first got connection refused. A socket table from a running guest with IPv6 showed every client of four loopback services connected over `::1`, so relaying those ports needs the IPv6 loopback too.

`bind_loopback_listeners()` binds each port on `127.0.0.1` and `[::1]`. Both listeners relay to the same gateway port. A guest that cannot bind `::1` logs that and keeps the IPv4 listener.

## Test results

```
$ make test-agent-unit FILTER=forward_localhost
# before the fix, with the helper binding 127.0.0.1 only
  FAIL forward_localhost_listens_on_both_loopback_addresses
  assertion `left == right` failed: a ::1 listener must exist exactly when this environment can bind ::1: [127.0.0.1:44431]
# after
  PASS forward_localhost_listens_on_both_loopback_addresses
  Summary: 1 test run: 1 passed, 173 skipped

$ make lint
exit 0
```

Not yet run end to end in a VM. The first such run will be a guest whose services connect over `::1`.

## Summary by CodeRabbit

* **New Features**
  * Localhost forwarding now supports connections over both IPv4 (`127.0.0.1`) and IPv6 (`::1`) loopback addresses.

* **Bug Fixes**
  * IPv4 forwarding remains available when IPv6 binding is unavailable.

* **Documentation**
  * Updated forwarding documentation to clarify IPv4 and IPv6 localhost support.


ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's review body is its generated walkthrough of the change, and its summary comment on this PR says no actionable comments were generated. No finding was raised.

Issue #954, opened 2026-09-19
Restore acknowledgement waits on serial socket destroys and three ip forks


ejc3:
On a guest with many TCP connections, the guest side of a restore spends most of its time in two steps that are serial by construction: destroying the snapshot-time sockets one netlink round trip at a time, and reconfiguring IPv6 with three `ip` processes. Both sit before the restore acknowledgement, and the guest's services cannot start reconnecting until the socket cleanup ends.

## Measured

A 128 GiB guest with about 850 established TCP connections at snapshot time, routed networking, 12 restores on one host. From the `restored` records' `guest_phases`:

| phase | range |
|---|---|
| `ipv6_ms` | 2,534 to 9,710 |
| `tcp_cleanup_ms` | 5,720 to 11,750 |
| of which `tcp_destroy_ms` | 3,976 to 9,399 |
| `total_ms` | 10,409 to 24,978 |
| host-observed `ack_ms` | 17,480 to 32,039 |

The low ends are copy-mode restores with the recorded working set replayed; the high ends are minor-mode restores. So part of the cost is page faults on cold guest pages, but 4 to 9 s for about 850 destroys is 5 to 11 ms per socket, and each destroy is one `SOCK_DESTROY` request followed by a wait for its acknowledgement (`fc-agent/src/snapshot_network.rs`).

`network::reconfigure_ipv6` (`fc-agent/src/network.rs`) runs `ip -6 addr show`, `ip -6 addr del` and `ip -6 addr add` as three child processes. After a restore every fork and exec of a binary that was cold at snapshot time faults its pages in through the memory server.

## What a fix could do

- Send the `SOCK_DESTROY` requests for the whole manifest on one netlink socket and then collect the acknowledgements, keeping the per-cookie verification.
- Do the IPv6 address swap over netlink from fc-agent itself, with no child processes.

Either change needs a restore benchmark before and after on a guest with hundreds of connections; the phases above are already in the `restored` record.

ejc3:
[...]

PR #979, opened 2026-09-21
fc-agent: retire the snapshot socket manifest over one netlink socket

ejc3:
Part of #954: the restore acknowledgement waits for fc-agent to destroy every snapshot-time socket, and it did that with a netlink socket per snapshot socket.

## The change

The cleanup opens one `NETLINK_SOCK_DIAG` socket for the whole manifest and sends the `SOCK_DESTROY` requests 64 to a datagram. Each acknowledgement is matched back to its socket by sequence number, and the per-cookie verification is unchanged: only an `NLMSG_ERROR` that echoes `SOCK_DESTROY` with that socket's sequence number decides it (0 destroyed, `ENOENT` or `ESTALE` already gone). Any other reply fails the restore closed and names the socket.

A batch has to fit the receive buffer, because an acknowledgement that does not fit is dropped and the next receive fails with `ENOBUFS`. Before the first batch the destroyer asks for 128 KiB with `SO_RCVBUFFORCE` (`SO_RCVBUF` without `CAP_NET_ADMIN`), reads back what the kernel granted, and sends smaller batches when that holds fewer than 64 acknowledgements at 2 KiB each.

Two behaviours differ from the per-socket loop: the 30 s cleanup budget is checked before each batch rather than before each socket, and when one acknowledgement in a batch reports a failure, the kernel has already handled the rest of that batch. They are all snapshot-time sockets, and the restore fails closed either way.

## Measured

8-vCPU, 8 GiB routed guest holding 800 connections to a listener inside the guest (about 1,600 sockets in the manifest), UFFD copy-mode restores, arms alternated, medians of 5 restores per arm after a warm-up, two sessions:

```
tcp_destroy_ms   158.5 -> 91.4    153.2 -> 103.6
tcp_cleanup_ms   253.0 -> 192.6   236.3 -> 208.8
total_ms         386.7 -> 324.9   393.7 -> 328.4
host ack_ms      475 -> 414       461 -> 391
```

## Tests

`make test-agent-unit` (178 passed at this commit), `make lint`, `make build`. Reds watched before the fix:

- `a_large_manifest_is_retired_over_one_socket_in_bounded_batches` and `acknowledgements_are_matched_to_their_socket_by_sequence_number` failed against a mutant that opened a socket per cookie and matched replies by arrival order.
- `a_lowered_receive_buffer_drops_no_acknowledgement` failed before the buffer reservation with "No buffer space available (os error 105)".

## Summary by CodeRabbit

- **Performance**
  - Snapshot restoration now retires network sockets in batches, improving efficiency when restoring configurations with many sockets.
  - Batch sizing adapts to the available receive buffer for more reliable processing.

- **Reliability**
  - Socket restoration accurately tracks individual outcomes, including sockets that have already disappeared.
  - Unsupported or failed socket operations now fail safely instead of being treated as successful.


ejc3 (commented):
NOT-A-DEFECT: CodeRabbit's review body is its generated walkthrough of the change, and its summary comment on this PR says no actionable comments were generated. No finding was raised.

PR #980, opened 2026-09-21
fc-agent: swap IPv6 and flush neighbours over rtnetlink on restore

ejc3:
Part of #954. **Stacked on:** fc-agent-sockdiag-batches (PR #979).

Before its restore acknowledgement, fc-agent started five `ip` processes: three for the IPv6 swap and two for the neighbour flush and pin. In a restored guest, every fork and exec of a binary that was cold at snapshot time faults its pages in through the memory server. fc-agent now sends the same requests itself over `NETLINK_ROUTE`, so the restore path starts no `ip` process before the acknowledgement.

## The change

- **IPv6 swap:** an `RTM_GETADDR` dump finds the first global-scope IPv6 address on eth0, `RTM_DELADDR` removes it, and `RTM_NEWADDR` adds the clone's /128 with `IFA_F_NODAD`. Failures are warnings as before, and a failed lookup or delete still adds the new address.
- **Neighbour flush and pin:** eth0's neighbours of both families are dumped and every entry `ip neigh flush` selects is deleted, for at most ten rounds. PERMANENT and NOARP entries and other interfaces are untouched. A delete that fails with ENOENT counts the entry as already gone, because the kernel's neighbour garbage collection can remove it between the dump and the delete; any other error fails the flush. The pin is the `RTM_NEWNEIGH` with create, replace and `NUD_PERMANENT` that `ip neigh replace ... nud permanent` sends.
- **Wire format:** every request matches what the guest's iproute2 6.1 sent in a restored clone, captured with strace, and unit tests compare against those bytes.
- **Shared framing:** the netlink framing moves to `fc-agent/src/netlink.rs`. Both dumps read through one `read_dump` with the interruption policy as a parameter: the socket manifest refuses an interrupted dump, and the address lookup keeps it with a warning, as iproute2 does.
- **Bounds:** each receive waits at most the socket's five-second timeout, where the child processes had no bound.

## Measured

Same guest and host as the base PR, arms alternated, medians of 5 restores per arm after a warm-up:

```
                                    before this PR -> after
ipv6_ms (IPv6 swap commit only)       87.0 -> 26.3
host ack_ms (whole PR)                 413 -> 381   (370-446 -> 329-403)
total_ms                             334.5 -> 306.6 (305-368 -> 270-340)
between IPv6 and TCP cleanup          97.7 -> 62.9  (83-113 -> 50-71)
```

The IPv6 commit alone did not move the acknowledgement, because the next cold `ip` exec paid the cost it saved; the neighbour commit removes that exec. After every restore eth0 held the clone's /128 and link-local address and the same neighbour entries in both arms.

The processes still on the path before the acknowledgement are the TCP gate's: `iptables -S` and `ip6tables -S` to verify it (about 20 ms), and one `iptables-restore` per family to remove it (about 65 ms). With NFS shares mapped, the remount also runs `umount`.

## Tests

`make test-agent-unit` (187 passed at the top), `make lint`, `make build` at each commit. Reds watched: the three address tests failed against a mutant that ignored the address scope and dropped `IFA_F_NODAD`; `an_interrupted_dump_is_refused_or_kept_as_asked` failed against a `read_dump` that checked the flag only on `NLMSG_DONE`; the five neighbour tests failed against stubs that sent nothing; `an_entry_collected_before_its_delete_does_not_fail_the_flush` failed before the ENOENT handling with "a neighbour delete failed: No such file or directory (os error 2)".

## Summary by CodeRabbit

* **Performance**
  * Network configuration during clone restoration now communicates directly with the kernel, reducing reliance on external commands.
* **Reliability**
  * Neighbour cleanup, health-check pinning, and IPv6 reconfiguration now complete before dependent networking steps continue.
  * Network updates and snapshot socket diagnostics provide stricter validation and improved handling of interrupted operations.
* **Tests**
  * Added coverage for IPv6 address swapping, neighbour management, and netlink dump behavior.


ejc3:
@coderabbitai review

ejc3 (commented):
RED-VERIFIED: a_stalled_ipv6_swap_leaves_the_runtime_worker_free (in 1b667ce2) and stalled_neighbour_steps_leave_the_runtime_worker_free (in 83eea311). Each runs the restore step on a one-worker runtime against a model netlink socket whose first receive stalls for one second, and requires another spawned task to run within 500 ms. Both failed first with "another task could not run while the restore step waited on the kernel". The IPv6 swap, the neighbour flush and the pin now run on tokio's blocking pool, and the restore awaits them in the same order.

ejc3:
@coderabbitai review

ejc3 (commented):
NOT-A-DEFECT: CodeRabbit refreshed its walkthrough comment when it answered the 16:44 review request, and the refreshed comment carries no finding. Its one finding on this PR, the netlink calls blocking the async worker, was answered RED-VERIFIED at 14:18 by a_stalled_ipv6_swap_leaves_the_runtime_worker_free and stalled_neighbour_steps_leave_the_runtime_worker_free, and its review of 83eea311 at 14:17 reported no actionable comments.

